PR #1, opened 2020-08-26
Removing the extra layer in the repo tree

bdematte:
Due to the last commit, convention copy is adding an extra repository which is breaking the project.
call to `basedir` is removing this extra repository.

### DIFF
--- a/boilerplate/update
+++ b/boilerplate/update
@@ -103,7 +103,7 @@ while read directory junk; do
   fi
   rm -rf "${DIR}/${directory}"
   mkdir -p $(dirname "${DIR}/${directory}")
-  rsync -a -r "$dir_path" "${DIR}/${directory}"
+  rsync -a -r "$dir_path" $(dirname "${DIR}/${directory}")
   if [ -f "${DIR}/${directory}/update" ]; then
     echo "executing ${DIR}/${directory}/update POST"
     "${DIR}/${directory}/update" POST


README.md
```diff
@@ -182,3 +182,8 @@ In your fork of this repository (not a consuming repository):
       `update` driver and the convention subdirectories themselves. Of
       note, `${CONVENTION_ROOT}/_lib/` contains some utilities that may
       be useful for `update`s.
+
+To test your changes, you can use `BOILERPLATE_GIT_REPO` environment 
+variable and set it to you local clone in order to overide the version of 
+boilerplate used. As `update` is cloning the git repo, ensure to have your 
+changes locally committed for your testing.
```
2uasimojo, line 189: Couple of typos:

```suggestion
To test your changes, you can use the `BOILERPLATE_GIT_REPO` environment 
variable and set it to your local clone in order to overide the version of 
boilerplate used. As `update` is cloning the git repo, ensure you have your 
changes locally committed for your testing.
```

There's another variable, `BOILERPLATE_GIT_CLONE`, that you would use if you wanted to e.g. add flags -- or even, for local testing, use a totally different command (like `rsync` :P). Do you think it's worth documenting that?

Also, an example would be great. Especially one that clones from my local machine, as I don't know how to do that :P

2uasimojo:
For now I'll have to take your word for it that this fix is needed, and approve so we can move on. But it would be great to get this validated via the test framework.

/lgtm

bdematte:
Just rebased on master to force triggering the build
/lgtm

PR #3, opened 2020-08-26
Switch to golangci-lint for the gocheck target

arjunrn:
Replace all the linting tooling with [golangci-lint](https://github.com/golangci/golangci-lint) which is a wrapper for the major linting tools. The changes also add a script which ensures that the linting tool is present on the path before doing the actual linting.


boilerplate/openshift/golang_osd_cluster_operator/ensure.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
2uasimojo, line 1: This is a cool idea. It would be neat to see as much of it as possible made part of [_lib](https://github.com/openshift/boilerplate/tree/master/boilerplate/_lib) and just have the configuration (some kind of declarative list of dep/ver/loc) in the convention subdir in a file with a known name.

2uasimojo: Gah, ignore this comment. I was thinking of this completely wrong. Your intent was (correctly) to run this as part of the `make` target, not to run it during `update`.

boilerplate/openshift/golang_osd_cluster_operator/standard.mk
```diff
@@ -71,8 +71,8 @@ docker-push: push
 
 .PHONY: gocheck
 gocheck: ## Lint code
-	boilerplate/openshift/golang_osd_cluster_operator/golint.sh
-	${GOENV} go vet ./cmd/... ./pkg/...
+	boilerplate/openshift/golang_osd_cluster_operator/ensure.sh golangci-lint
```
2uasimojo, line 74: Yeah, I missed this line before

boilerplate/openshift/golang_osd_cluster_operator/standard.mk
```diff
@@ -71,8 +71,8 @@ docker-push: push
...
+	GOLANGCI_LINT_CACHE=/tmp/golanci-cache golangci-lint run -c boilerplate/openshift/golang_osd_cluster_operator/golangci.yml ./...
```
2uasimojo, line 75: nit:
```suggestion
	GOLANGCI_LINT_CACHE=/tmp/golangci-cache golangci-lint run -c boilerplate/openshift/golang_osd_cluster_operator/golangci.yml ./...
```
I'm curious how this env var is used by golangci-lint. If you don't set it, does it use a default? (I couldn't find documentation for it at a quick glance.)

arjunrn: I'm not sure why it resolves to `/.cache` as the default directory for caching but this [issue](https://github.com/golangci/golangci-lint/issues/734) summarizes what's going on.

2uasimojo:
/lgtm

Do we want to wait to merge this until one or more of the following is in place?
- At least one or two guinea pig repositories, using the old thing, to keep the initial bootstrap PR smaller
- In-repo testing framework
- Automated merge

If not, that's fine, just a thought.

boilerplate/_lib/ensure.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+set -euo pipefail
+
+GOLANGCI_LINT_VERSION="1.30.0"
+DEPENDENCY=${1:-}
+
+case "${DEPENDENCY}" in
+    golangci-lint)
```
2uasimojo, line 8: wouldn't hurt to have a default case that blows up if you pass a dep we don't understand (or misspell one we do).

boilerplate/openshift/golang_osd_cluster_operator/standard.mk
```diff
@@ -71,8 +71,8 @@ docker-push: push
 
 .PHONY: gocheck
 gocheck: ## Lint code
-	boilerplate/openshift/golang_osd_cluster_operator/golint.sh
-	${GOENV} go vet ./cmd/... ./pkg/...
+	boilerplate/_lib/ensure.sh golangci-lint
+	GOLANGCI_LINT_CACHE=/tmp/golangci-cache golangci-lint run -c boilerplate/openshift/golang_osd_cluster_operator/golangci.yml ./...
```
2uasimojo, line 75: Could put a comment here explaining why we need `GOLANGCI_LINT_CACHE`, pointing to that issue.

2uasimojo:
Thanks for the updates @arjunrn 

/lgtm again

2uasimojo:
Apparently gh picks and chooses which directives are allowed to have stuff on the same line.

/lgtm

PR #6, opened 2020-08-26
Add no-op `make pr-check`

2uasimojo:
...for PROW. This will be fleshed out later when we have something to
check.


2uasimojo:
Used #5 instead.

PR #11, opened 2020-08-27
[WIP] - First version of the compare function for testing framework

bdematte:
Implementing comparison function in order to ensure the content after update matches the content from the convention.
It also contains the start of a new testing convention in order to integrate testing cases without impacting the actual conventions.


test/lib.sh
```diff
@@ -59,3 +59,59 @@ EOF
...
+	declare -i rc
+	
+	pushd $repo/boilerplate > /dev/null
+	
+	compare _data
+	compare _lib
+	
+	while read convention ; do
+	  if [ -d $REPO_ROOT/boilerplate/$convention ] ; then
+		  compare $convention
```
2uasimojo, line 103: This needs to set `rc` to nonzero if `compare` fails.

test/case/01_bootstrap_update
```diff
@@ -18,3 +18,18 @@ cd $repo
...
 make update_boilerplate
+
+check_update 
+
+if [ $? -ne 0 ] ; then
+	return $?
+fi
+
+add_convention openshift/test_base_convention
+add_convention openshift/test_base
```
2uasimojo, line 29: A nonexistent convention should cause `update` to fail. The output looks promising:

```
Invalid convention directory: 'openshift/test_base'
make[1]: *** [Makefile:3: update_boilerplate] Error 1
make[1]: Leaving directory '/tmp/tmp.fOXGsNq9ug'
```

...but the overall test case isn't failing. We need to fix that.

[Later] There are a couple of pieces causing this. One is that the main driver needs to *not* use `source` [here](https://github.com/openshift/boilerplate/pull/11/files#diff-ad6c39a41f5fcfb2b3d89656bfb00755R28). I saw that yesterday but forgot to take it out. Another is that we need to capture errors [here](https://github.com/openshift/boilerplate/pull/11/files#diff-e1b2b21859e872e9af54745ed0a1b5c4R103) -- commented accordingly.

test/lib.sh
```diff
@@ -59,3 +59,59 @@ EOF
 hr() {
     echo "========================="
 }
+
+diff() {
+	for file in `ls` ; do 
+		if [ -d $file ] ; then 
+			pushd $file  > /dev/null
+			diff $1/$file
```
2uasimojo, line 67: Did you mean to check `$?` here?

(On a broader note, the more I use bash's `-e`, the less I like it. I would rather we were instead explicit about all our error conditions.)

test/lib.sh
```diff
@@ -59,3 +59,59 @@ EOF
...
+		fi
+	done
+	return 0
+}
+
+
+compare() {
+	if [ $1 = "_data" ] ; then
+		if [ ! -f $repo/boilerplate/_data/last_boilerplate_commit ] ; then
+			rc=`expr $rc+1`
```
2uasimojo, line 83: Need to initialize `rc`. Unless you're counting on it coming down from the caller -- which IMO is not a good look.

Also, just set it to a hardcoded value rather than incrementing it, per [above comment](https://github.com/openshift/boilerplate/pull/11/files#diff-e1b2b21859e872e9af54745ed0a1b5c4R76). I know in this case it can only be hit once, but good habits.

test/lib.sh
```diff
@@ -59,3 +59,59 @@ EOF
...
+	popd > /dev/null
+	
+	return $rc
+}
+
+add_convention() {
+	if [ $(grep $1\$ $repo/boilerplate/update.cfg | wc -l) == 0 ] ; then
+		echo $1 >> $repo/boilerplate/update.cfg
+	fi
+	make update_boilerplate
```
2uasimojo, line 116: We should avoid running `make update_boilerplate` in these helper methods, instead letting the calling environment do it explicitly. For example, at some point we'll want to write a test that adds multiple conventions at once before updating.

test/driver
```diff
@@ -15,7 +15,7 @@ _FAIL_COUNT=0
 main() {
     cd "$HERE/case"
     # TODO: Add support for running just certain cases
-    cases=$(find . -type f -perm /111 | sort)
+    cases=$(find . -type f -perm +111 | sort)
```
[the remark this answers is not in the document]
bdematte: Yes, it is not working on Mac, would have been too easy. Will switch to the 'string' syntax tomorrow then, which should work on all systems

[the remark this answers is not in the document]
2uasimojo: Made zero test cases fail via #14.

test/case/01_bootstrap_update
2uasimojo: Fixed via #14.

test/lib.sh
2uasimojo: Removed `-e` from test case 01 (and avoided it in the new 02) via #14.

test/lib.sh
bdematte: If I am not mistaken, the `declare -i rc` (line 107) is initialising it to 0. The objective is to use to rc to return how many checks failed basically, hence the init in the check_method.

test/lib.sh
```diff
@@ -59,3 +59,59 @@ EOF
...
+			diff $1/$file 
+			popd  > /dev/null
+		elif [ -f $file ] ; then
+			cmp $file $1/$file
+			if [ ! $? = 0 ] ; then
+				rc=`expr $rc+1`
+			fi
+		fi
+	done
+	return 0
```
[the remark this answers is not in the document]
bdematte: In fact, it is left-over from the initial mocking, I need to review the return in the different pieces indeed :D

test/driver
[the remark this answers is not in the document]
2uasimojo: > Yes, it is not working on Mac, would have been too easy.

I'm starting to get concerned that you're using some archaic version of standard shell tools on your Mac.

test/lib.sh
2uasimojo: Okay, right, so to expand on my "not a good look" comment, it's not good practice for the two functions to be so tightly coupled. In particular, if I wanted to use `compare` from somewhere else, I would have to be sure to set up `rc` beforehand. That's weird -- and also not really necessary.

If you want to count failures from `compare`, you can increment a variable local to the caller _from_ the caller. (But we still shouldn't use that count as a return code, for reasons stated.)

test/driver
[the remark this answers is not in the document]
bdematte: Jaein. It is archaic in the sense MacOS is using BSD flavored binary, which is not supporting the '/' .
There is an option to install GNU versions but I am not sure we should put such constraints if we can avoid.

test/lib.sh
```diff
@@ -59,3 +59,59 @@ EOF
...
+}
+
+
+compare() {
+	if [ $1 = "_data" ] ; then
+		if [ ! -f $repo/boilerplate/_data/last_boilerplate_commit ] ; then
```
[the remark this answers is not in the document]
bdematte: Yes. that was my intent

test/lib.sh
bdematte: I implemented a dedicated function for init/increment/check the counter actually.

PR #12, opened 2020-08-27
Try to capture console output

2uasimojo:


2uasimojo:
Notes to self:

In order to make this work, we would have to get that variable passed in from the job itself, e.g. by making [this line](https://github.com/openshift/release/blob/fdee01f83f75f60b69df809a7dcd164682f96475/ci-operator/config/openshift/boilerplate/openshift-boilerplate-master.yaml#L15) do something like:

```yaml
commands: OUTPUT_DIR=${ARTIFACT_DIR} make pr-check
```

and having `make pr-check` use `$OUTPUT_DIR` to manually redirect output.

That seems silly. The job prints the console output when it fails; why can't it (at least optionally based on a setting) also print it when it succeeds?

Anyway, dropping this for now.

PR #14, opened 2020-08-27
Harden test framework

2uasimojo:
- Fail if no test cases found.
- Run, don't source, the test cases.
- Improve test driver summary output.
- Make test case 01_bootstrap_update explicit about errors (don't use `-e`).
- Add test case 02_nonexistent_convention, proving that subscribing to a nonexistent convention causes the update to fail.

### DIFF
--- a/test/case/02_nonexistent_convention
+++ b/test/case/02_nonexistent_convention
@@ -1,0 +1,43 @@
+#!/bin/bash -x
+
+echo "Validate that a nonexistent convention causes update to fail."
+
+HERE=${0%/*}
+
+REPO_ROOT=$(git rev-parse --show-toplevel)
+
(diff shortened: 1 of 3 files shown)


test/case/02_nonexistent_convention
```diff
@@ -0,0 +1,43 @@
...
+source $REPO_ROOT/test/lib.sh
+
+repo=$(empty_repo)
+add_cleanup $repo
+
+bootstrap_repo $repo
+
+cd $repo
+# Subscribe to a "convention" that doesn't exist upstream
+echo "bogus/convention" >> boilerplate/update.cfg
```
2uasimojo, line 18: This can use `add_convention` from #11 once that lands.

PR #15, opened 2020-08-28
Update lib.sh to integrate compare part

bdematte:
Split of [PR#11](https://github.com/openshift/boilerplate/pull/11)
This contains only the updates of lib.sh for the comparison including : 
- Diffs count is now being done within a dedicated function to avoid shared variable.
- Make update_boilerplate has been removed from add_convention to delegate it to test case

### DIFF
--- a/test/lib.sh
+++ b/test/lib.sh
@@ -15,7 +15,7 @@ _cleanup() {
     if [ -z "$PRESERVE_TEMP_DIRS" ]; then
         echo "Removing temporary directories"
         rm -fr $_BP_TEST_TEMP_DIRS
-		rm -rf $LOG_DIR
+        rm -rf $LOG_DIR
     else
         echo "Preserving temporary directories: $_BP_TEST_TEMP_DIRS $LOG_DIR"
     fi
@@ -52,7 +52,7 @@ bootstrap_repo() {
         cat <<EOF > Makefile
 .PHONY: update_boilerplate
 update_boilerplate:
-	@boilerplate/update
+		@boilerplate/update
 EOF
         > boilerplate/update.cfg
     )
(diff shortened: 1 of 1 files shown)


test/lib.sh
```diff
@@ -59,3 +59,98 @@ EOF
...
+# specific parts
+check_update() {
+	handle_error_counter INIT
+	pushd $repo/boilerplate > /dev/null
+	
+	compare _data
+	compare _lib
+	
+	while read convention ; do
+	  if [ -d $BOILERPLATE_GIT_REPO/boilerplate/$convention ] ; then
```
2uasimojo, line 113: I feel like either we don't need this condition, or it should have an `else blow up` clause.

If there's a convention in the config that's not in boilerplate, `update` itself will fail. But if it doesn't, we need to know about it. So I prefer adding the `else`.

test/lib.sh
```diff
@@ -59,3 +59,98 @@ EOF
...
+compare() {
+	if [ $1 = "_data" ] ; then
+		if [ ! -f $repo/boilerplate/_data/last_boilerplate_commit ] ; then
+			# TODO: Check the content of the file to ensure it contains the proper commit in addition to the file existence
+			handle_error_counter INC
+		fi
+	else
+		pushd $1  > /dev/null
+		diff $BOILERPLATE_GIT_REPO/boilerplate/$1
+		popd > /dev/null
```
2uasimojo, line 97: As noted above, why can't this just be

```suggestion
		diff --recursive $BOILERPLATE_GIT_REPO/boilerplate/$1 $1 || handle_error_counter INC
```

...where by `diff` I mean the `diff` program; and we don't need the `diff` function.

test/lib.sh
```diff
@@ -59,3 +59,98 @@ EOF
...
+	  if [ -d $BOILERPLATE_GIT_REPO/boilerplate/$convention ] ; then
+		  compare $convention
+	  fi
+	done < $repo/boilerplate/update.cfg
+	
+	popd > /dev/null
+	
+	handle_error_counter CHECK
+	
+	exit $?
```
2uasimojo, line 122: This should be
```suggestion
	return $?
```
Let the caller decide whether it's time to `exit`.

test/lib.sh
```diff
@@ -59,3 +59,98 @@ EOF
...
+handle_error_counter() {
+	declare -i error_counter
+	if [ $1 = "RESET" ] ; then 
+		error_counter=0
+	elif [ $1 = "INC" ] ; then
+		error_counter=`expr $error_counter+1`
+	elif [ $1 = "CHECK" ] ; then 
+		if [ ! $error_counter = 0 ] ; then 
+			echo "$error_counter differences have been detected between the project and the convention"
+			exit 1
```
2uasimojo, line 151:
```suggestion
			return 1
```

test/lib.sh
```diff
@@ -59,3 +59,98 @@ EOF
...
+		fi
+	fi
+	
+	exit 0
```
2uasimojo, line 155:
```suggestion
	return 0
```

test/lib.sh
```diff
@@ -59,3 +59,98 @@ EOF
...
+
+## handle_error_counter FUNCTION
+#
+# Function managing the error counter to increment in case of diff found for a file and print before returning
+# :param FUNCTION: action to be done on the error counter
+#  - RESET : Initilize the error counter to 0 
+#  - INC : Increment the internal counter by 1
+#  - CHECK : Check the number of diff is 0. If it is not, returns 1 and print the number of errors found
+handle_error_counter() {
+	declare -i error_counter
```
2uasimojo, line 143: This is declaring a local variable. It goes out of scope as soon as the function returns. So after fixing the `exit`s below:

```
[efried@efried boilerplate]$ handle_error_counter RESET
[efried@efried boilerplate]$ handle_error_counter INC
[efried@efried boilerplate]$ handle_error_counter CHECK
[efried@efried boilerplate]$ echo $?
0
```

test/lib.sh
```diff
@@ -59,3 +59,98 @@ EOF
...
+## diff BASELINE
+#
+# Recursively compare the current repository content to the BASELINE repository
+#
+# :param BASELINE: An existing directory used as reference to compared sync'ed files
+diff() {
+	for file in `ls` ; do 
+		if [ -d $file ] ; then 
+			pushd $file  > /dev/null
+			diff $1/$file
```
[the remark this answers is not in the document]
bdematte: because `diff --recursive` would not allow to know how many files differs and I believe this is meaningful information for debugging. After, we can say we don't want/need this information but I think it would be helpful (and eventually, add the list of different files in the check).

test/lib.sh
```diff
@@ -59,3 +59,98 @@ EOF
...
+#  - RESET : Initilize the error counter to 0 
+#  - INC : Increment the internal counter by 1
+#  - CHECK : Check the number of diff is 0. If it is not, returns 1 and print the number of errors found
+handle_error_counter() {
+	declare -i error_counter
+	if [ $1 = "RESET" ] ; then 
+		error_counter=0
+	elif [ $1 = "INC" ] ; then
+		error_counter=`expr $error_counter+1`
+	elif [ $1 = "CHECK" ] ; then
```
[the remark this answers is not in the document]
bdematte: I hesitated as there will be extra conditions shortly (on the number of parameter for the function by example). Do we prefer `case` with inner `if` or a single `if` ?

test/lib.sh
bdematte: going through the `man`, looks like `-q` would provide the needed details (number of diff and which diff). Will update the script

test/lib.sh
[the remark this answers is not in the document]
bdematte: checking around, this can be worked around with the `-q` option and having/generating a log file to have the list

test/lib.sh
bdematte: Added an entry in the diff output which will result in failing test

test/lib.sh
```diff
@@ -59,3 +61,72 @@ EOF
...
+
+## compare FOLDER LOG_FILE
+#
+# Check FOLDER is properly sync'ed, determining the reference base on FOLDER 
+#
+# :param FOLDER: An existing directory sync'ed by boilerplate and to be checked
+# :param LOG_FILE: The log file used to aggregate the output of the `diff` calls
+compare() {
+	if [ $1 = "_data" ] ; then
+		if [ ! -f $repo/boilerplate/_data/last_boilerplate_commit ] ; then
```
2uasimojo, line 73: Sorry, I just noticed you've used this `$repo` variable in several places throughout this file, but I don't see anywhere it's defined.

2uasimojo: Okay, I see it's being set in the test case itself. Needs to be a parameter instead.

test/lib.sh
```diff
@@ -60,97 +62,71 @@ hr() {
...
 	pushd $repo/boilerplate > /dev/null
 	
-	compare _data
-	compare _lib
+	if [ $# = 1 ] ; then
+		LOG_FILE=$LOG_DIR/$1
+		rm -f $LOG_FILE
+	else 
+		log_file=`cat /dev/urandom | env LC_CTYPE=C tr -cd 'a-f0-9' | head -c 10`
+		LOG_FILE=$LOG_DIR/$log_file
```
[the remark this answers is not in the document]
bdematte: this is to create log file, not a dir (as there is already a dir for logs, didn't want to add unneeded layers)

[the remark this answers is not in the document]
2uasimojo: Right, `mktemp` creates a temporary file if you don't specify `-d`.

test/lib.sh
```diff
@@ -50,7 +52,7 @@ bootstrap_repo() {
         cat <<EOF > Makefile
 .PHONY: update_boilerplate
 update_boilerplate:
-	@boilerplate/update
+		@boilerplate/update
```
2uasimojo, line 55: random whitespace damage

2uasimojo:
For reasons unknown, some repositories accept `/unhold`, but some don't. I think it's spelled

/hold cancel

around here {shrug}.

PR #16, opened 2020-08-28
Update test 01 to add content check 

bdematte:
Updating the test case 01 to include the comparison of the sync'ed content

### DIFF
--- a/boilerplate/test/test_base_convention/.codecov.yml
+++ b/boilerplate/test/test_base_convention/.codecov.yml
@@ -1,0 +1,26 @@
+codecov:
+  notify:
+    require_ci_to_pass: no
+
+coverage:
+  precision: 2
+  round: down
+  range: "20...100"
+
+  status:
+    project: no
+    patch: no
+    changes: no
+
+parsers:
+  gcov:
+    branch_detection:
+      conditional: yes
+      loop: yes
+      method: no
+      macro: no
+
+comment:
+  layout: "reach,diff,flags,tree"
+  behavior: default
+  require_changes: no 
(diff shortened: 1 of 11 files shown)


bdematte:
This has a dependency on https://github.com/openshift/boilerplate/pull/15

boilerplate/test/test_base_convention/.codecov.yml
```diff
@@ -0,0 +1,26 @@
+codecov:
```
2uasimojo, line 1: Still don't want the files in test conventions to be real files.

test/case/01_bootstrap_update
```diff
@@ -18,4 +18,16 @@ cd $repo
...
+check_update 
+
+if [ $? -ne 0 ] ; then
+	exit $?
+fi
+
+add_convention test/test_base_convention
+make update_boilerplate
+
+check_update
```
2uasimojo, line 31: I tried to get this to fail by adding a line here that modifies one of the convention files before `check_update`. It didn't work. I didn't dig in.

boilerplate/test/test_base_convention/.codecov.yml
bdematte: Even if they don't have to be real life ones, is there any specific issue you are thinking of for having to differ ?

2uasimojo: Sorry, I'm not quite parsing your question.

I'm just complaining that it's confusing to see real content in these files. For example, somebody in the future is going to think they need to keep them in sync with what's in the real conventions.

It's really not a big deal though.

2uasimojo:
I ran this through a few times, injecting various failures, and verified it works as expected. Let's merge this now and worry about making the content of the test convention less meaningful in a followup PR.

/lgtm
/approve

PR #18, opened 2020-08-31
Also build go binary before image build

arjunrn:
The `build` target builds the docker image for the operators. But it may not work because the binary required to build the image is missing. This adds `gobuild` as a prerequisite for the docker image build.

### DIFF
--- a/boilerplate/openshift/golang_osd_cluster_operator/standard.mk
+++ b/boilerplate/openshift/golang_osd_cluster_operator/standard.mk
@@ -54,7 +54,7 @@ isclean:
 	@(test "$(ALLOW_DIRTY_CHECKOUT)" != "false" || test 0 -eq $$(git status --porcelain | wc -l)) || (echo "Local git checkout is not clean, commit changes and try again." >&2 && exit 1)
 
 .PHONY: build
-build: isclean envtest
+build: isclean envtest gobuild
 	docker build . -f $(OPERATOR_DOCKERFILE) -t $(OPERATOR_IMAGE_URI)
 	docker tag $(OPERATOR_IMAGE_URI) $(OPERATOR_IMAGE_URI_LATEST)
 


2uasimojo:
It's hard to see because we don't have the Dockerfile in this repo yet, but the docker build triggers `gobuild` in the container. So I don't believe we want to do this.

/hold

arjunrn:
@2uasimojo Yep a multistage docker build would work. But the repository I was testing against doesn't use it. It expects the binary to be present in the working directory. Should we assume that every repository will just use multistage builds?

2uasimojo:
> @2uasimojo Yep a multistage docker build would work. But the repository I was testing against doesn't use it. It expects the binary to be present in the working directory. Should we assume that every repository will just use multistage builds?

I'm afraid I don't understand what you mean by "multistage", or which "binary" you're referring to.

I'm quite inexperienced when it comes to docker. But what I can tell you is that, in the SREP-owned repositories I've looked at, the pattern seems to be:
- `make gobuild` does the go build
- Dockerfile includes running `make gobuild`
- `make build` runs the docker build.
- `hack/app_sre_build_deploy.sh` runs `make build` (and also a couple other things like push)

...so this is the pattern I assumed we would be adopting for boilerplate. (Plus moving `hack/app_sre_build_deploy.sh` into the bp payload, and triggering it via a `make build-deploy` target from app-interface rather than having the path hardcoded there.)

Abstractly, we definitely need a way to ensure that the go build takes place in a container we control. We mustn't rely on whatever environment happens to be present on the CI server. (The same precept applies to the other check/build bits and pieces, as we've been discussing, like the [YAML validation](https://github.com/openshift/boilerplate/pull/19).)

arjunrn:
Right, so in the repos you're seen in there are 2 stages. The first stage of the docker build is done with one image and the entire git repo is copied into it and built there, with `make gobuild`. Subsequently another image is created and the operator binary built in the previous stage is copied into the new image. But this is not how all images are built. It's generally more common for the compiled binary to be built on the host and then copied into the image. In that case you need both the docker build target to first build the go binary. But let's just go with the multistage builds for now.

PR #19, opened 2020-08-31
Added yaml validation as a phony target

arjunrn:
Added a make target to validate yaml files. This target is also a requisite for the `test` target. By default it checks for YAML files in the path `deploy/crds`. If the path doesn't exist then it does nothing. Otherwise it goes through the directory and validates every YAML file. The location of the YAML files can be overriden by setting the environment variable `CRDS_PATH` with the path to the directory or file.

### DIFF
--- a/boilerplate/openshift/golang_osd_cluster_operator/standard.mk
+++ b/boilerplate/openshift/golang_osd_cluster_operator/standard.mk
@@ -28,6 +28,8 @@ OPERATOR_DOCKERFILE ?=build/Dockerfile
 BINFILE=build/_output/bin/$(OPERATOR_NAME)
 MAINPACKAGE=./cmd/manager
 
+CRDS_PATH ?= deploy/crds
+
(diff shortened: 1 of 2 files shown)


boilerplate/openshift/golang_osd_cluster_operator/standard.mk
```diff
@@ -28,6 +28,8 @@ OPERATOR_DOCKERFILE ?=build/Dockerfile
 BINFILE=build/_output/bin/$(OPERATOR_NAME)
 MAINPACKAGE=./cmd/manager
 
+CRDS_PATH ?= deploy/crds
```
2uasimojo, line 31: Maybe we want to discover instead. Something like

```
git ls-files | grep -v ^vendor/ | egrep '\.ya?ml'
```

(The python script is set up to accept files or directories.)

2uasimojo:
Yeah, nice.

Would you please write up a card to address running all checks/builds in containers?

/lgtm
/approve

PR #22, opened 2020-09-01
Ensure golangci-lint for the correct OS

### LINKED ISSUE #20: Dependency Installation Broken For golangci-lint
The golangci-lint is broken. I'm using `gvm` and on a mac.

```
boilerplate/_lib/ensure.sh golangci-lint
tar: could not chdir to '/Users/dustinrow/.gvm/pkgsets/go1.13.6/global/bin'

make: *** [gocheck] Error 1
```

arjunrn:
Attempts to fix #20 

Currently the `ensure.sh` attempts to download the `golangci-lint` tool even when the tool is present on the path. The `ensure.sh` will now download the script only when the tool is not present in `$PATH`. Also other minor improvements:

- Ensure that `$GOPATH/bin` is present in `$PATH`.
- Allow user to specify the path for `$GOLANGCI_LINT_CACHE`

### DIFF
--- a/boilerplate/_lib/ensure.sh
+++ b/boilerplate/_lib/ensure.sh
@@ -5,15 +5,24 @@ GOLANGCI_LINT_VERSION="1.30.0"
 DEPENDENCY=${1:-}
 
 case "${DEPENDENCY}" in
-    golangci-lint)
-        GOPATH=$(go env GOPATH)
-        if [ ! -f "${GOPATH}/bin/golangci-lint" ]; then
-            DOWNLOAD_URL="https://github.com/golangci/golangci-lint/releases/download/v${GOLANGCI_LINT_VERSION}/golangci-lint-${GOLANGCI_LINT_VERSION}-linux-amd64.tar.gz"
-            curl -sfL "${DOWNLOAD_URL}" | tar -C "${GOPATH}/bin" -zx --strip-components=1 "golangci-lint-${GOLANGCI_LINT_VERSION}-linux-amd64/golangci-lint"
+golangci-lint)
+    GOPATH=$(go env GOPATH)
+    if which golangci-lint ; then
+        exit
+    else
+        mkdir -p "${GOPATH}/bin"
+        echo "${PATH}" | grep -q "${GOPATH}/bin"
+        IN_PATH=$?
+        if [ $IN_PATH != 0 ]; then
+            echo "${GOPATH}/bin not in $$PATH"
+            exit 1
         fi
+        DOWNLOAD_URL="https://github.com/golangci/golangci-lint/releases/download/v${GOLANGCI_LINT_VERSION}/golangci-lint-${GOLANGCI_LINT_VERSION}-${GOOS}-amd64.tar.gz"
+        curl -sfL "${DOWNLOAD_URL}" | tar -C "${GOPATH}/bin" -zx --strip-components=1 "golangci-lint-${GOLANGCI_LINT_VERSION}-${GOOS}-amd64/golangci-lint"
+    fi
     ;;
-    *)
-        echo "Unknown dependency: ${DEPENDENCY}"
-        exit 1
+*)
+    echo "Unknown dependency: ${DEPENDENCY}"
+    exit 1
     ;;
 esac
--- a/boilerplate/_lib/ensure.sh
+++ b/boilerplate/_lib/ensure.sh
@@ -3,6 +3,7 @@ set -euo pipefail
 
 GOLANGCI_LINT_VERSION="1.30.0"
 DEPENDENCY=${1:-}
+GOOS=${GOOS:-linux}
 
 case "${DEPENDENCY}" in
 golangci-lint)
--- a/boilerplate/_lib/ensure.sh
+++ b/boilerplate/_lib/ensure.sh
@@ -3,7 +3,7 @@ set -euo pipefail
 
 GOLANGCI_LINT_VERSION="1.30.0"
 DEPENDENCY=${1:-}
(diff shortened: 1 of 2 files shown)


boilerplate/_lib/ensure.sh
```diff
@@ -5,15 +5,24 @@ GOLANGCI_LINT_VERSION="1.30.0"
...
+    else
+        mkdir -p "${GOPATH}/bin"
+        echo "${PATH}" | grep -q "${GOPATH}/bin"
+        IN_PATH=$?
+        if [ $IN_PATH != 0 ]; then
+            echo "${GOPATH}/bin not in $$PATH"
+            exit 1
         fi
+        DOWNLOAD_URL="https://github.com/golangci/golangci-lint/releases/download/v${GOLANGCI_LINT_VERSION}/golangci-lint-${GOLANGCI_LINT_VERSION}-${GOOS}-amd64.tar.gz"
+        curl -sfL "${DOWNLOAD_URL}" | tar -C "${GOPATH}/bin" -zx --strip-components=1 "golangci-lint-${GOLANGCI_LINT_VERSION}-${GOOS}-amd64/golangci-lint"
```
dustman9000, line 21: Ensure that `GOOS` is defined

dustman9000:
This is failing locally on my machine:

```
$ make
boilerplate/_lib/ensure.sh golangci-lint
# GOLANGCI_LINT_CACHE needs to be set to a directory which is writeable
# Relevant issue - https://github.com/golangci/golangci-lint/issues/734
GOLANGCI_LINT_CACHE=/tmp/golangci-cache golangci-lint run -c boilerplate/openshift/golang_osd_cluster_operator/golangci.yml ./...
bash: /Users/dustinrow/.gvm/pkgsets/go1.13.6/global/bin/golangci-lint: cannot execute binary file
make: *** [gocheck] Error 126
dustinrow@drow-mac:~/src/must-gather-operator$ which golangci-lint
/Users/dustinrow/.gvm/pkgsets/go1.13.6/global/bin/golangci-lint
```

arjunrn:
You will have 

> I tested this by moving `golangci-lint` out of my $PATH and it's failing locally for me:

You will have to set and pass the `GOOS` environment variable when calling `make`. Otherwise it defaults to `linux` as you suggested.

boilerplate/_lib/ensure.sh
```diff
@@ -3,17 +3,27 @@ set -euo pipefail
 
 GOLANGCI_LINT_VERSION="1.30.0"
 DEPENDENCY=${1:-}
+GOOS=${GOOS:-linux}
```
dustman9000, line 6: You should really get the host OS via `go env GOOS` instead of checking the variable as it will fail to execute if the downloaded file doesn't match.

dustman9000:
Tested locally with latest changes and it pulls down the proper golangci-lint regardless of cross-compile GOOS.

boilerplate/_lib/ensure.sh
```diff
@@ -3,17 +3,27 @@ set -euo pipefail
...
 case "${DEPENDENCY}" in
-    golangci-lint)
-        GOPATH=$(go env GOPATH)
-        if [ ! -f "${GOPATH}/bin/golangci-lint" ]; then
-            DOWNLOAD_URL="https://github.com/golangci/golangci-lint/releases/download/v${GOLANGCI_LINT_VERSION}/golangci-lint-${GOLANGCI_LINT_VERSION}-linux-amd64.tar.gz"
-            curl -sfL "${DOWNLOAD_URL}" | tar -C "${GOPATH}/bin" -zx --strip-components=1 "golangci-lint-${GOLANGCI_LINT_VERSION}-linux-amd64/golangci-lint"
+golangci-lint)
+    GOPATH=$(go env GOPATH)
+    if which golangci-lint ; then
+        exit
```
2uasimojo, line 12: This doesn't ensure we're using `$GOLANGCI_LINT_VERSION`. As written, there are several ways that could go awry even when we're running the `gocheck` target in a container -- e.g. if the base container image we're using already has the binary (at the wrong version).

So IMO we really ought to include a version check here. Something like:
```suggestion
    if which golangci-lint && [[ $(golangci-lint --version) == *" v$GOLANGCI_LINT_VERSION "* ]]; then
        exit
```

boilerplate/_lib/ensure.sh
```diff
@@ -3,17 +3,27 @@ set -euo pipefail
...
+    else
+        mkdir -p "${GOPATH}/bin"
+        echo "${PATH}" | grep -q "${GOPATH}/bin"
+        IN_PATH=$?
+        if [ $IN_PATH != 0 ]; then
+            echo "${GOPATH}/bin not in $$PATH"
+            exit 1
         fi
+        DOWNLOAD_URL="https://github.com/golangci/golangci-lint/releases/download/v${GOLANGCI_LINT_VERSION}/golangci-lint-${GOLANGCI_LINT_VERSION}-${GOOS}-amd64.tar.gz"
+        curl -sfL "${DOWNLOAD_URL}" | tar -C "${GOPATH}/bin" -zx --strip-components=1 "golangci-lint-${GOLANGCI_LINT_VERSION}-${GOOS}-amd64/golangci-lint"
```
2uasimojo, line 22: Doesn't the `amd64` bit also break on Mac? (`ppc` or something?)

arjunrn: No Macs have used intel x86 architecture for about 10 years now.

2uasimojo: TIL.

We're still totally screwing all the people trying to develop on AIX though.

PR #24, opened 2020-09-02
Switch to git submodule

NautiluX:
Switch from copy and pasting across repos to submodules for
bootstrapping.

This should help
* improve bootstrapping experience
* track the changes from boilerplate itself
* make self-updates easier as new versions are just checked out
* Users can easily switch back to a previous version of boilerplate in case of
issues


NautiluX:
/close
:sadpanda:

PR #25, opened 2020-09-02
$CONVENTION_NAME

2uasimojo:
When boilerplate/_lib/common.sh is sourced by a convention `update`
script, it's useful for that script to know the name of its convention
without having to hardcode it. This commit adds logic to common.sh to
discover this and put it in a `$CONVENTION_NAME` variable. The
openshift/golang_osd_cluster_operator convention's `update` script is
modified to use the new variable.

As part of this discovery, `$HERE` (the directory containing the file
doing the sourcing) is also discovered and set. This commit removes it
from places where it is no longer needed (or never was).

Also remove a redundant shebang from a lib that should only ever be
sourced.

### DIFF
--- a/boilerplate/_lib/common.sh
+++ b/boilerplate/_lib/common.sh
@@ -3,3 +3,29 @@ err() {
   exit 1
 }
 
+# Only used for error messages
+_lib=${BASH_SOURCE##*/}
+
+# When this lib is sourced (which is what it's designed for), $0 is the
+# script that did the sourcing.
+SOURCER=$(realpath $0)
+[[ -n "$SOURCER" ]] || err "$_lib couldn't discover where it was sourced from"
+
+HERE=${SOURCER%/*}
+[[ -n "$HERE" ]] || err "$_lib failed to discover the dirname of sourcer at $SOURCER"
+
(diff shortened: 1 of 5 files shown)


boilerplate/_lib/common.sh
```diff
@@ -1,3 +1,17 @@
+# When this lib is sourced (which is what it's designed for), $0 is the
+# script that did the sourcing.
+SOURCER=$(realpath $0)
+HERE=${SOURCER%/*}
+
+REPO_ROOT=$(git rev-parse --show-toplevel)
+CONVENTION_ROOT=$REPO_ROOT/boilerplate
+
+# Were we sourced from within a convention?
+if [[ "$HERE" == "$CONVENTION_ROOT/"* ]]; then
```
dustman9000, line 10: Should we have more defensive checks around `$HERE` and `$CONVENTION_ROOT`? Seems like if these are empty we should `exit 1`

2uasimojo: Can do. (Though we would really have to be trying to make either of those things happen.)

jharrington22 (approved):
/lgtm

PR #27, opened 2020-09-02
Copy golang_osd_cluster_operator to ..._hive_...

2uasimojo:
At this point in history, all the content of the
openshift/golang_osd_cluster_operator convention is actually applicable
to both cluster-installed and hive-installed operators. So there should
be nothing stopping us from publishing both conventions.

Eventually we're going to have a smarter way to maintain the common bits
in a single place. But for now we're dumbly copying the contents of
golang_osd_cluster_operator into golang_osd_hive_operator.

While these conventions are duplicates of each other, it's important
that they stay in sync. So this commit also adds a test case to enforce
that. The test case is expected to be temporary, to be removed as soon
as the aforementioned "smarter way" is implemented.

### DIFF
--- a/boilerplate/openshift/golang_osd_hive_operator/.codecov.yml
+++ b/boilerplate/openshift/golang_osd_hive_operator/.codecov.yml
@@ -1,0 +1,26 @@
+codecov:
+  notify:
+    require_ci_to_pass: no
+
+coverage:
+  precision: 2
+  round: down
+  range: "20...100"
+
+  status:
+    project: no
+    patch: no
+    changes: no
+
+parsers:
+  gcov:
+    branch_detection:
+      conditional: yes
+      loop: yes
+      method: no
+      macro: no
+
+comment:
+  layout: "reach,diff,flags,tree"
+  behavior: default
+  require_changes: no 
--- a/boilerplate/openshift/golang_osd_hive_operator/codecov.sh
+++ b/boilerplate/openshift/golang_osd_hive_operator/codecov.sh
@@ -1,0 +1,40 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+REPO_ROOT=$(git rev-parse --show-toplevel)
+CI_SERVER_URL=https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test
+COVER_PROFILE=${COVER_PROFILE:-coverage.out}
+JOB_TYPE=${JOB_TYPE:-"local"}
+
+make -C "${REPO_ROOT}" gotest TESTOPTS="-coverprofile=${COVER_PROFILE}.tmp -covermode=atomic -coverpkg=./..."
+
+# Remove generated files from coverage profile
+grep -v "zz_generated" "${COVER_PROFILE}.tmp" > "${COVER_PROFILE}"
+rm -f "${COVER_PROFILE}.tmp"
+
+# Configure the git refs and job link based on how the job was triggered via prow
+if [[ "${JOB_TYPE}" == "presubmit" ]]; then
+       echo "detected PR code coverage job for #${PULL_NUMBER}"
+       REF_FLAGS="-P ${PULL_NUMBER} -C ${PULL_PULL_SHA}"
+       JOB_LINK="${CI_SERVER_URL}/pr-logs/pull/${REPO_OWNER}_${REPO_NAME}/${PULL_NUMBER}/${JOB_NAME}/${BUILD_ID}"
+elif [[ "${JOB_TYPE}" == "postsubmit" ]]; then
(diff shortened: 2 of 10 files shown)


boilerplate/openshift/golang_osd_hive_operator/includes.mk
```diff
@@ -0,0 +1,4 @@
+# This is a rollup of includes in this convention, for easier inclusion
+# by the repo's Makefile, which should just include this file.
+include boilerplate/openshift/golang_osd_cluster_operator/project.mk
```
dustman9000, line 3: Should be `openshift/golang_osd_hive_operator` right?

boilerplate/openshift/golang_osd_hive_operator/includes.mk
```diff
@@ -0,0 +1,4 @@
...
+include boilerplate/openshift/golang_osd_cluster_operator/standard.mk
```
dustman9000, line 4: Same here

boilerplate/openshift/golang_osd_hive_operator/.codecov.yml
```diff
@@ -0,0 +1,26 @@
+codecov:
```
dustman9000, line 1: Can we DRY this up? I hate to see copied files like this. I was hoping for putting all of these common files into an `openshift/golang_osd_operator_common` convention or something.

boilerplate/openshift/golang_osd_hive_operator/codecov.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
```
dustman9000, line 1: Same comment on DRY'ing this up into a common convention for operators.

boilerplate/openshift/golang_osd_hive_operator/includes.mk
2uasimojo: Yeah.

This made me go off and implement #28, in which this file goes away.

It was only when I was done with it that I realized I would need to address this issue anyway, because the path is also hardcoded into standard.mk. So

/hold

What I'll actually want to do is have standard.mk discover the path and set a variable, in a fashion similar to what #25 did. That way it'll work wherever we drop it.

boilerplate/openshift/golang_osd_hive_operator/includes.mk
2uasimojo: Will fix, see above.

boilerplate/openshift/golang_osd_hive_operator/.codecov.yml
2uasimojo: We've talked around this extensively over the past couple of days and come up with some great ideas for commonizing, dependency management, etc.

Those things are eminently doable, but a nontrivial amount of extra work that we would like to defer for the time being.

We also talked about it early on in the design process and decided the same thing.

The maintenance is pretty much `cp -r`. The test lets you know when you missed doing that.

boilerplate/openshift/golang_osd_hive_operator/codecov.sh
2uasimojo: See above.

2uasimojo:
So I'm going to rebase this on #28 and at the same time fix standard.mk to do dynamic discovery of its own subdirectory.

bdematte:
Any reason not to go for https://issues.redhat.com/browse/OSD-5008 straight ? 
The test to diff the content of convention is fairly unmaintainable (even short term) and implementation cost of OSD-5008 isn't higher than this solution.

2uasimojo:
> Any reason not to go for https://issues.redhat.com/browse/OSD-5008 straight ?

I talked that through at length with @arjunrn yesterday. The devil, as always, is in the details. I think the main objection to that implementation was the softness of the dependency management, and requiring the consumer to subscribe to two conventions, one of which they'll eventually have to unsubscribe from.

No matter which way we slice it, we're going to have to have something hackish in place until we go full bore and implement the ultimate solution. I thought this was an okay compromise. But I can be swayed.
 
> The test to diff the content of convention is fairly unmaintainable (even short term)

I don't agree with this. The only way that diff changes is if we actually start adding hive- or cluster-specific pieces, and we're quite a long way from that. We can cut over to a better solution (better than OSD-5008) by then.

2uasimojo:
In today's sync we agreed to implement [OSD-5008](https://issues.redhat.com/browse/OSD-5008)-ish (I'll update that card accordingly):
- Collapse (well, retain) everything in a single convention
- Name it `golang_osd_operator` (but change everything to "kebab case", so => `golang-osd-operator`)
- Defer designing a full scale dependency structure etc. until we actually start needing it.

2uasimojo:
Superseded by #31

PR #28, opened 2020-09-03
One-time "nexus" Makefile include

2uasimojo:
Previously, any Makefile content in subscribed conventions needed to be
pulled in -- and maintained on each update if the filenames or locations
changed -- by manually curating `include` directives in your base
Makefile.

With this commit, you only need to add

`include boilerplate/generated-includes.mk`

to your base Makefile after you bootstrap (or update to/past this
commit). The contents of that file are regenerated on each update to
automatically `include` any `*.mk` files in your subscribed conventions.


boilerplate/update
```diff
@@ -10,6 +10,9 @@
 # script.
 
 set -e
+if [ "$BOILERPLATE_SET_X" ]; then
```
arjunrn, line 13: I would recommend calling this `DEBUG` or `BP_DEBUG` and using this as a convention in all scripts.

2uasimojo: > and using this as a convention in all scripts.

To be clear: By putting it in the two places I've put it, it automatically applies to any `make update_boilerplate` or `make test` all the way through. We don't need to add it to individual test cases or anything.

> I would recommend calling this `DEBUG` or `BP_DEBUG`

I can rename it if you like.

2uasimojo (commented):
Was there anything else of concern here @arjunrn ?

boilerplate/openshift/golang_osd_cluster_operator/update
```diff
@@ -17,7 +17,7 @@ THINGS YOU NEED TO DO
 =====================
 - Make sure the following line is in your base Makefile:
 
-include boilerplate/$CONVENTION_NAME/includes.mk
+include boilerplate/generated-includes.mk
```
dustman9000, line 20: I'm not a big fan of a single entry point like this. You're assuming the only way to consume a conventions is through Make. I'd much rather keep the entry points in each convention subdirectory and keep the boilerplate directory clean. We then let the consumers explicitly include them.

boilerplate/update
arjunrn: I meant if there are any other shell scripts if they all use the same environment variable for debug logs then it would be quite convenient. Because the `-x` flag won't be passed to any shell scripts which are called from this script. But we can discuss that later separately.

arjunrn:
> Was there anything else of concern here @arjunrn ?

Nope.
/lgtm

boilerplate/openshift/golang_osd_cluster_operator/update
2uasimojo: @dustman9000 Sorry this merged before we got a chance to finish this discussion. I'll try to summarize what we spoke about in person earlier:

> I'm not a big fan of a single entry point like this. You're assuming the only way to consume a conventions is through Make.

The only thing this assumes is that you'll consume *make targets* through Make. It doesn't affect the disposition of any boilerplate files, or any other aspect of the framework.

> I'd much rather keep the entry points in each convention subdirectory.

The advantage of the approach in this PR is that the consumer doesn't have to curate the includes in their base Makefile every time they subscribe/unsubscribe to/from a convention, or (more importantly) on updates in case the convention has changed. Yes, we have READMEs, and yes we print instructions from the per-convention `update` scripts, but this way we don't have to rely on a) not messing those up, and b) people actually reading them.

> We then let the consumers explicitly include them.

We'd like to discourage the ability for consumers to e.g. include some but not all of the `*.mk`s -- accidentally (as described above) or deliberately.

> I'd much rather {snip} keep the boilerplate directory clean.

Agreed. We discussed ways to reorganize the boilerplate directory to make it cleaner. For example:

- `boilerplate/infra/` to contain things like the `update` script, `update.cfg`, the hash of the last update, and this `generated-includes.mk` file.
- `boilerplate/conventions` to contain actual convention subdirectory structures.

That's pretty much orthogonal to this PR, but in any case something we can certainly do later.

PR #29, opened 2020-09-03
Switch to kebab-case

2uasimojo:
Use kebab-case for file names and make targets.


test/case/03-test-update-from-master
```diff
@@ -21,43 +21,43 @@ LOG=${0##*/}.log
 boilerplate_master=$(new_boilerplate_repo)
```
2uasimojo, line 21: NOTE TO REVIEWERS: I changed the mode on this file to disable this test case. Because it straddles master and $current_commit, it was going to be harder-than-reasonable to make it work with the renamed convention, and we're going to need to undo whatever accommodation we implement anyway. This was the easiest thing for now.

test/case/03-test-update-from-master
```diff
@@ -21,43 +21,43 @@ LOG=${0##*/}.log
...
 
-make update_boilerplate >$LOG 2>&1
+make update-boilerplate >$LOG 2>&1
 rc=$?
 cat $LOG
 if [ $rc -ne 0 ] ; then
 	exit $rc
 fi
-check_update $repo 03_check_after_init
+check_update $repo 03-check-after-init
```
dustman9000, line 32: Should `check_update` be `check-update`?

test/lib.sh
```diff
@@ -35,7 +35,7 @@ add_cleanup() {
 }
 
 empty_repo() {
```
dustman9000, line 37: Not sure if `bash` supports hyphens, but should we change these to kebab-case?

test/case/03-test-update-from-master
2uasimojo: This is a bash function. I didn't rename any of those. (Is that even legal? /me tries... Yup. Still, not planning to rename all the bash functions everywhere, that okay?)

test/lib.sh
2uasimojo: As above: it does (but I had to try it to find out), but IMO no we shouldn't.

test/case/03-test-update-from-master
2uasimojo: Re-enabled via #30.

PR #31, opened 2020-09-03
One convention to rule them all

2uasimojo:
Since all the current boilerplate is common to both cluster- and
hive-deployed operators, stop propagating separate conventions for
those. Instead move all the content into a single, common
`openshift/golang-osd-operator` convention.

It will be up to future us to decide how to maintain the common pieces
once we do have divergent content for different kinds of consumers.

Jira: [OSD-5008](https://issues.redhat.com/browse/OSD-5008)


arjunrn:
/lgtm
/hold

Please unhold when all reviews are done.

2uasimojo:
Thanks for the reviews.

/hold cancel

PR #32, opened 2020-09-04
Adding checks on generated code 

bdematte:
Adding new make target which ensure the version of generated code which is committed is up-to-date.
On top of this, creating a dedicated tree in the `test` repository validating the convention features (separated from the core testing of boilerplate).


test/case/convention_testing/openshift/golang-osd-cluster-operator/01_generated_files_checker
```diff
@@ -0,0 +1,49 @@
...
+	echo "No error detected while checking on uncommitted project"
+	exit 1
+else
+	echo "Project is not committed. Checks failed as expected"
+fi
+
+git add -A
+git commit -m "initial commit"
+
+make check_generation
```
arjunrn, line 31: Not a comment, more of a question. I don't understand why things should fail here. Because the modified changes are just been committed in the previous line.

bdematte: The project initially does not contain the generated files so after the check_generation, as it trigger the code generation, there is a new diff.

arjunrn: I would recommend that the test for the gogenerate not only test that the file has been generated but also that the contents of the generated file match the expected generated out. Also a test case can be added where the contents of the generated file are modified and detected.

2uasimojo (commented):
I didn't get all the way through this.

It'll need a rebase, as some of the files were moved via #31.

Speaking of renames: I like that you moved the test cases into a framework-specific subdirectory. Nit: can we just call it `framework` instead of `framework_testing`? The "testing" part is redundant: it's already under the `test/` subdirectory of the repo root.

boilerplate/openshift/golang-osd-cluster-operator/standard.mk
```diff
@@ -110,7 +110,14 @@ envtest: isclean
 	@eval $$($(MAKE) env --no-print-directory) || (echo 'Unable to evaulate output of `make env`.  This breaks osd-operators-registry.' >&2 && exit 1)
 
 .PHONY: test
-test: envtest gotest yaml-validate
+test: envtest gotest yaml-validate check_generation
+
+.PHONY: check_generation
+check_generation: 
+	@$(MAKE) -s isclean --no-print-directory || (echo 'check_generation requires a clean project to run. Please commit or revert you changes first.' >&2 && exit 1) 
+	@$(MAKE) -s gogenerate --no-print-directory
```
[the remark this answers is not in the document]
bdematte: As far as I remember, there is no ordering in the deps as they can be ran in parallel, meaning we can't have `gogenerate` as a deps. And because of that, I thought it would be more consistent to have all done the same way and with appropriate error message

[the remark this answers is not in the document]
bdematte: Yes, I had to downgrade it for now in the last commit as it was breaking the testing and I wanted to validate it was only the sdk part which was failing. Will work on this today

boilerplate/update
```diff
@@ -125,7 +125,8 @@ while read directory junk; do
   fi
   echo "adding makefile includes."
   for inc in $(find "${DIR}/${directory}" -type f -name '*.mk' | sort); do
-    echo "include ${inc#$REPO_ROOT/}" >> "${NEXUS_MK}"
+    real_include=`realpath $inc`
```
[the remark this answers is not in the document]
bdematte: `git remote get-url` returns a path with resolved symlinks, so this is failing in case your `TMPDIR` contains a symlink

boilerplate/openshift/golang-osd-cluster-operator/standard.mk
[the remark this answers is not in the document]
2uasimojo: > As far as I remember, there is no ordering in the deps as they can be ran in parallel

Right you are. Withdrawn.

boilerplate/update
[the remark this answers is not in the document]
2uasimojo: I don't understand the symlink thing.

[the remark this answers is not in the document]
2uasimojo: Or perhaps it was the `git remote get-url` thing that was throwing me off. That's not in play here. Perhaps you mean `find` is not resolving symlinks by default? If that's the issue, we should use `find -L` rather than trying to hack it up after the fact.

test/case/convention/openshift/golang-osd-operator/01_generated_files_checker
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash -x
```
2uasimojo, line 1: Remove `-x`. Set `BOILERPLATE_SET_X` to enable it on a per-run basis.

Also, use `-e` as you convinced me we should :P

test/case/convention/openshift/golang-osd-operator/01_generated_files_checker
```diff
@@ -0,0 +1,51 @@
...
+
+source $REPO_ROOT/test/lib.sh
+
+repo=$(empty_repo)
+add_cleanup $repo
+test_project="file-generate"
+
+# TODO: test REPO_NAME detection by setting an appropriately-formatted
+# git remote
+export REPO_NAME=${0##*/}
```
2uasimojo, line 13: Keep an eye on #33 and remove this once it lands.

test/case/convention/openshift/golang-osd-operator/01_generated_files_checker
```diff
@@ -0,0 +1,51 @@
...
+fi
+
+git add -A
+git commit -m "initial commit"
+
+echo $GOPATH
+
+make check-generate
+if [ $? = 0 ] ; then 
+    echo "Error not detected while the generated files have not beent added/committed"
```
2uasimojo, line 35: Is this message backwards?
```suggestion
    echo "Error not detected while the generated files have been added/committed"
```

test/case/convention/openshift/golang-osd-operator/01_generated_files_checker
```diff
@@ -0,0 +1,51 @@
...
+    exit 1
+else
+    echo "Generated files are not committed. Checks failed as expected"
```
2uasimojo, line 38:
```suggestion
    echo "Generated files are committed. Checks succeeded as expected"
```
?

test/case/convention/openshift/golang-osd-operator/01_generated_files_checker
```diff
@@ -0,0 +1,51 @@
...
+
+if [ ! -f src/test/test-generation ] || [ ! -f src/test/subfolder/test-generation-subfolder ] ; then
+    echo "Generation didn't work properly"
+    exit 1
+fi
+
+git add -A
+git commit -m "commit generated files"
+
+make check-generate
```
2uasimojo, line 49: Redundant?

test/lib.sh
```diff
@@ -44,6 +44,31 @@ empty_repo() {
...
 }
 
+## bootstrap_project PATH TEST_PROJECT INITIAL_CONVENTION
+#
+# Build a temp boilerplated git project containing test_project files
+# - Copies in boilerplate/update from $REPO_ROOT
+# - Creates an empty boilerplate/update.cfg
+# It does not run the update.
+#
+# :param PATH: the test_project (from boilerplate/test/test_projects)
```
2uasimojo, line 54: Add the other two params

test/lib.sh
```diff
@@ -44,6 +44,31 @@ empty_repo() {
...
+    (
+        cp -R $REPO_ROOT/test/test_projects/$test_project/* $repodir/.
+        pushd $repodir > /dev/null
+        mkdir boilerplate
+        cp $REPO_ROOT/boilerplate/update boilerplate
+        touch boilerplate/update.cfg
+        for convention in $3 ; do
+            add_convention . $convention
+        done
+        sh boilerplate/update
```
2uasimojo, line 67: Belies L52

test/lib.sh
```diff
@@ -44,6 +44,31 @@ empty_repo() {
...
+# - Creates an empty boilerplate/update.cfg
+# It does not run the update.
+#
+# :param PATH: the test_project (from boilerplate/test/test_projects)
+bootstrap_project() {
+    repodir=$1
+    test_project=$2
+    (
+        cp -R $REPO_ROOT/test/test_projects/$test_project/* $repodir/.
+        pushd $repodir > /dev/null
```
2uasimojo, line 60: Prefer not swallowing stdio unless there's a good reason for it. E.g. when we're relying on a function to print a specific thing for the caller to consume. But even then, prefer to work around by using `>&2` so the output is still visible.

test/lib.sh
```diff
@@ -44,6 +44,31 @@ empty_repo() {
...
+        mkdir boilerplate
+        cp $REPO_ROOT/boilerplate/update boilerplate
+        touch boilerplate/update.cfg
+        for convention in $3 ; do
+            add_convention . $convention
+        done
+        sh boilerplate/update
+        popd $repodir/$test_project > /dev/null
```
2uasimojo, line 68: ditto stdios

test/test_projects/file-generate/Makefile
```diff
@@ -0,0 +1,6 @@
+include boilerplate/generated-includes.mk
+
+.PHONY: update_boilerplate
+update_boilerplate:
+	@boilerplate/update
```
2uasimojo, line 5: kebab-case please

test/test_projects/file-generate/.gitignore
```diff
@@ -0,0 +1,77 @@
+# Temporary Build Files
```
2uasimojo, line 1: Can we please name this directory `test/convention/...`?

boilerplate/update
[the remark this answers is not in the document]
2uasimojo: Summarizing offline explanation:

When we calculate `$REPO_ROOT` on [L64](https://github.com/openshift/boilerplate/pull/32/files#diff-fb7297e9f79b9a50001fa66af778f797R64), `git rev-parse` returns the result fully resolved. But `$DIR` is calculated using `pwd` ([L17](https://github.com/openshift/boilerplate/pull/32/files#diff-fb7297e9f79b9a50001fa66af778f797R17)) which doesn't fully resolve. So if the path contains symlinks, the remove-left-pattern below won't match.

Notwithstanding the weirdness of @bdematte's system having a symlink from `/var` to `private/var` (/me rolls eyes at Mac *again*) I _guess_ this makes it important that both are fully resolved.

Arguably it would be better to do it at L17 instead, e.g.

```bash
DIR=$(realpath ${0%/*})
```

test/case/convention/openshift/golang-osd-operator/01_generated_files_checker
bdematte: No. Initial commit does not contain the files to be generated (as the is_clean check is failling, file generation is not triggered by the first `check-generate`). 
As a consequence, this step is generating new files which are not committed in git, and so, the check is expected to fail due to the existing diff.
The 3rd check (after the commit including the generated files, l.47 ) is testing the check is passing when everything is generated/committed.

test/test_projects/file-generate/.gitignore
bdematte: This is not containing convention but actual 'light' project using conventions, not sure it would be meaningful to call it `convention`.

test/case/convention/openshift/golang-osd-operator/01_generated_files_checker
2uasimojo: Right, got it, thanks.

test/case/convention/openshift/golang-osd-operator/01_generated_files_checker
2uasimojo: See above, this is legit.

test/test_projects/file-generate/.gitignore
2uasimojo: What was in my head was: this is used to test (a component of) the `golang-osd-operator` convention. But take your point that it's a "test project". Can we just call it `test/project/...` then? Again, I feel like `test_` is redundant since the subdir is already `test/`. (But if you do end up with a name with multiple words, please remember to use kebab-case.)

test/lib.sh
bdematte: Will update it to move forward but not sure if we should enforce blindly this logic or try to be relevant. 
The objective here is to make stdio readable/useful (avoiding to have context less repository names/list) and alligned with `cd` which is not polluting stdout.

test/case/convention_testing/openshift/golang-osd-cluster-operator/01_generated_files_checker
bdematte: Yes, that's planned once the 'simple' case is working.

PR #33, opened 2020-09-04
Test: convention update invocation and variables

2uasimojo:
This rewrites the `update` script in the `test/test-base-convention`
convention to test that the following documented behaviors of the main
update driver are observed when it is invoked:

- Exactly one command line argument, either `PRE` or `POST`.
- Exported variables `REPO_ROOT`, `REPO_NAME`, and `CONVENTION_ROOT`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,9 @@
 ALLOW_DIRTY_CHECKOUT?=false
 
+# Tests rely on this starting off unset. (And if it is set, it's usually
+# not for the reasons we care about.)
+unexport REPO_NAME
+
 .PHONY: isclean
 isclean:
 	@(test "$(ALLOW_DIRTY_CHECKOUT)" != "false" || test 0 -eq $$(git status --porcelain | wc -l)) || (echo "Local git checkout is not clean, commit changes and try again." >&2 && exit 1)
--- a/boilerplate/test/test-base-convention/update
+++ b/boilerplate/test/test-base-convention/update
@@ -1,29 +1,22 @@
 #!/bin/bash -e
 
-source $CONVENTION_ROOT/_lib/common.sh
+# This script validates the vars we expect from the main `update` driver, and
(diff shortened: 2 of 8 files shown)


2uasimojo:
Hmph, that's weird. Works like a charm locally. It's like the update script is being invoked from the wrong directory?

I'm going to turn on debug and see if we can get any more information.

/hold

2uasimojo:
Going to leave this to next week now, but investigation thus far has made things more, not less, confusing. Notes:

[Slack thread](https://coreos.slack.com/archives/CBN38N3MW/p1599252948086100) where @stevekuznetsov helped me understand how to get closer to reproducing what happens in the CI environment.

Steve and I were both able to observe a *third* behavior, `error: Unknown subcommand: get-url`, which seems to be happening because the container is running git 1.8.3.1. It is bizarre that this is different from the failure in the automated CI run; as far as we were able to tell, we were running the exact same command under the exact same container image. This is, like, the whole point of using containers.

Next step is probably to spin up a copy of the actual pod and run it in there rather than via `docker run`.

2uasimojo:
Okay, that last test may make things a little clearer. At least the git error is reproducible. Previously we weren't getting to that command because apparently the `$REPO_NAME` environment variable was already set in the pod environment (but not when we manually ran the container!?!).

So two things to unwind here:
- Figure out why `$REPO_NAME` is set or unset depending on automated vs. manual. And we may end up having to work around it somehow, which is unfortunate.
- Get a modern (like, come on, less than seven years old would be great) version of `git` in the container.

2uasimojo:
Conclusion:
- Don't care why `$REPO_NAME` is set in the automated CI environment. Unset it via Makefile.
- Accommodate primitive `git` versions by using a `git config` command instead of `git remote get-url`.
- Prioritize [OSD-5074](https://issues.redhat.com/browse/OSD-5074) so we can locally simulate what happens in the CI more closely.

PR #40, opened 2020-09-08
Ensure PyYAML is installed for yaml-validate

2uasimojo:
Eventually, we ought to ensure that boilerplate consumers all satisfy
their check/test/build targets via a standardized container that has
these things in it. But for now...


2uasimojo:
This is needed to make https://github.com/openshift/configure-alertmanager-operator/pull/101 work.

PR #41, opened 2020-09-09
Add check-generate target for go operator + workaround for missing operator-sdk in build

bdematte:
This is containing the same code than [PR#32](https://github.com/openshift/boilerplate/pull/32) with a workaround in the new test case to avoid build failure due to missing operator-sdk in the base image


boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -110,7 +110,14 @@ envtest: isclean
 	@eval $$($(MAKE) env --no-print-directory) || (echo 'Unable to evaulate output of `make env`.  This breaks osd-operators-registry.' >&2 && exit 1)
 
 .PHONY: test
-test: envtest gotest yaml-validate
+test: envtest gotest yaml-validate check-generate
+
+.PHONY: check-generate
+check-generate: 
+	@$(MAKE) -s isclean --no-print-directory || (echo 'check-generate requires a clean project to run. Please commit or revert you changes first.' >&2 && exit 1)
```
2uasimojo, line 117:
```suggestion
	@$(MAKE) -s isclean --no-print-directory || (echo 'check-generate requires a clean project to run. Please commit or revert your changes first.' >&2 && exit 1) 
```
...but doesn't the `isclean` target already print an appropriate message? I guess this one is (slightly) more specific to `check-generate`, but still seems kind of redundant.

Another thing: if I've got `ALLOW_DIRTY_CHECKOUT` set in my env, this will pass, and then the `generate` below will potentially end me up in an unrevertable state. I wonder if we should force `ALLOW_DIRTY_CHECKOUT=false` in this rule. Thoughts?

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -110,7 +110,14 @@ envtest: isclean
...
 
 .PHONY: test
-test: envtest gotest yaml-validate
+test: envtest gotest yaml-validate check-generate
+
+.PHONY: check-generate
+check-generate: 
+	@$(MAKE) -s isclean --no-print-directory || (echo 'check-generate requires a clean project to run. Please commit or revert you changes first.' >&2 && exit 1) 
+	@$(MAKE) -s generate --no-print-directory
+	@$(MAKE) -s isclean --no-print-directory || (echo 'Files after generation are different than committed one. Please commit updated and unaltered generated files' >&2 && exit 1)
```
2uasimojo, line 119:
```suggestion
	@$(MAKE) -s isclean --no-print-directory || (echo 'Files after generation are different than committed ones. Please commit updated and unaltered generated files' >&2 && exit 1)
```

test/case/convention/openshift/golang-osd-operator/01_generated_files_checker
```diff
@@ -0,0 +1,61 @@
...
+export PATH=`pwd`/bin:$PATH
+
+which operator-sdk
+
+add_cleanup $repo
+test_project="file-generate"
+
+# TODO: test REPO_NAME detection by setting an appropriately-formatted
+# git remote
+export REPO_NAME=${0##*/}
```
2uasimojo, line 39: This can go away now.

test/case/convention/openshift/golang-osd-operator/01_generated_files_checker
```diff
@@ -0,0 +1,61 @@
...
+        ;;
+esac
+curl -OJL https://github.com/operator-framework/operator-sdk/releases/download/${RELEASE_VERSION}/${binary}
+chmod +x ${binary}
+cp ${binary} bin/operator-sdk
+export PATH=`pwd`/bin:$PATH
+
+which operator-sdk
+
+add_cleanup $repo
```
2uasimojo, line 34: nit: let's keep this right after L7

test/lib.sh
```diff
@@ -44,6 +44,35 @@ empty_repo() {
...
+## bootstrap_project PATH TEST_PROJECT INITIAL_CONVENTION
+#
+# Build a temp boilerplated git project containing test_project files
+# - Copies in boilerplate/update from $REPO_ROOT
+# - Creates an empty boilerplate/update.cfg
+# It DOES run the update in order to allow using boilerplate/generated-includes.mk
+#
+# :param PATH: An existing directory that has been `git init`ed, like
+#       what you get when you run `empty_repo`.
+# :pranm TEST_PROJECT: the test_project (from boilerplate/test/test_projects)
```
2uasimojo, line 56:
```suggestion
# :param TEST_PROJECT: the test_project (from test/projects)
```

test/lib.sh
```diff
@@ -44,6 +44,35 @@ empty_repo() {
...
+# :param INITIAL_CONVENTION: The convention(s) to be used for initializing the 
+#       project
```
2uasimojo, line 58: nit: Could mention that, if specifying more than one convention, the argument needs to be quoted and whitespace-separated, like `"openshift/foo test/bar test/baz"`

test/lib.sh
```diff
@@ -44,6 +44,35 @@ empty_repo() {
...
+bootstrap_project() {
+    repodir=$1
+    test_project=$2
+    (
+        cp -R $REPO_ROOT/test/projects/$test_project/* $repodir/.
+        pushd $repodir
```
2uasimojo, line 64: nit: since you're running this chunk in a subshell (the parens on L62 & 73) you could just 
```suggestion
        cd $repodir
```
here, and...

test/lib.sh
```diff
@@ -44,6 +44,35 @@ empty_repo() {
...
+        mkdir boilerplate
+        cp $REPO_ROOT/boilerplate/update boilerplate
+        touch boilerplate/update.cfg
+        for convention in $3 ; do
+            add_convention . $convention
+        done
+        sh boilerplate/update
+        popd $repodir/$test_project
```
2uasimojo, line 72: ...nix this

test/projects/file-generate/.gitignore
```diff
@@ -0,0 +1,77 @@
+# Temporary Build Files
```
2uasimojo, line 1: I wouldn't mind a README.md in this "project" that explains what we're using it for. Including mentioning that the only piece that's actually important right now is the two test.go files containing `go:generate` directives.

2uasimojo (commented):
Nothing serious in here, but since you need to rebase anyway...

test/case/convention/openshift/golang-osd-operator/01_generated_files_checker
2uasimojo: D'oh, never mind, #33 isn't merged yet.

boilerplate/openshift/golang-osd-operator/standard.mk
bdematte: I thought the `-s` would silence the output, but only it silences only the targets names in fact. Will remove it indeed as not worth the redundancy.

test/case/convention/openshift/golang-osd-operator/01_generated_files_checker
2uasimojo: It is now :)

2uasimojo (changes requested):
Most of these comments are nits, mainly just fixing grammar. The only one that's actually important is the regression of #43.

The other thing that concerns me right now is that this puts operator-sdk in the path of `make test`. So consuming repositories will be forced to include operator-sdk in their build image, or their CI will break. I imagine most of them aren't doing that yet.

Boilerplate could pull it in (via `ensure.sh` for now; via our fancy build image eventually) -- but not until we're ready to force a specific version, and we're not there yet.

Then again, I suppose `ensure.sh` could parse go.mod to figure out the version to install. The following seems to work, properly taking `replace` directives into account:

```shell
go list -json -m github.com/operator-framework/operator-sdk | jq -r 'if .Replace != null then .Replace.Version else .Version end'
```

Run against aws-efs-operator, this outputs [`v0.16.0`](https://github.com/openshift/aws-efs-operator/blob/4bc180b792e71a32aa5b9cffdac43781fc041a51/go.mod#L10); run against gcp-project-operator it prints [`v0.8.2`](https://github.com/openshift/gcp-project-operator/blob/d7679b901911a59b9ad61ab43d5a79c679e5a8dc/go.mod#L74) (not [`v0.8.3-0.20190722210327-daf62d44e47e`](https://github.com/openshift/gcp-project-operator/blob/d7679b901911a59b9ad61ab43d5a79c679e5a8dc/go.mod#L31)).

This only works with modules, not with vendored deps, but I think we're probably okay with that.

2uasimojo:
I made #44 to do what I mentioned above: have ensure.sh install the correct version of operator-sdk based on what's in your go.mod. I stole some of your code from this PR -- which you should now be able to remove and just use directly from the convention.

bdematte:
> Boilerplate could pull it in (via `ensure.sh` for now; via our fancy build image eventually) -- but not until we're ready to force a specific version, and we're not there yet.
> 
> Then again, I suppose `ensure.sh` could parse go.mod to figure out the version to install. The following seems to work, properly taking `replace` directives into account

Issue is that `make generate` is tightly linked to the `operator sdk` version due to `operator-sdk` intergace instability (current target is not supported by operator-sdk 0.19 for example). For now, I would rather limit us to supporting 1 operator-sdk version. Else, we need to. : 
- implement a compatibility declaration for conventions so that we can ensure used `operator-sdk` from go.mod is supported by the convention
- update our testing to test each of the supported versions 

I don't think this would be needed for every single dependency but for operator-sdk (and knowing 1.0 will come with some/more impactful issue), this is needed if we want `auto-discovery`. 

Alternative would be to have 1 convention per operator-sdk version (with potential symlinks to avoid code dupplication) but this would not scale if/when we have a 2nd/3rd 'interface changing' dependency.

2uasimojo:
> Issue is that `make generate` is tightly linked to the `operator sdk` version due to `operator-sdk` intergace instability (current target is not supported by operator-sdk 0.19 for example).

Agree this is an issue. Discussing in #44. But...

> For now, I would rather limit us to supporting 1 operator-sdk version.

I like this as a long-term goal, but if we do it now, it'll choke adoption.

As mentioned in [#44](https://github.com/openshift/boilerplate/pull/44#discussion_r487076807), as long as nobody is at 0.19 or 1.x yet (I don't *think* they are??) we're really only talking about supporting two or three permutations across versions 0.8-0.18, of just this one target. Does that seem reasonable?

dustman9000:
> operator-sdk

Since we have familiarity with v0.16.0 and most of our operators are at that version, let's just settle on that version for now.

2uasimojo:
> Since we have familiarity with v0.16.0 and most of our operators are at that version, let's just settle on that version for now.

Sorry, I don't agree. I'm not sure what the official list of candidates is, but I'm pretty sure most of these qualify:

```
cloud-ingress-operator           v0.15.1
*configure-alertmanager-operator v0.16.0
deadmanssnitch-operator          v0.16.0
elasticsearch-operator           v0.18.1
*gcp-project-operator            v0.8.2
managed-upgrade-operator         v0.17.0
managed-velero-operator          v0.17.1
*must-gather-operator            v0.12.0
pagerduty-operator               v0.16.0
rbac-permissions-operator        v0.16.0
```
* current guinea pigs per [OSD-4943](https://issues.redhat.com/browse/OSD-4943)

I'm not overly concerned with upgrading gpo and mgo to v0.16.0 for the guinea pig round. What I *am* concerned about is trying to adopt muo and mvo (not sure whether elasticsearch-operator is in scope) which would need to be _downgraded_. We either need a way to accommodate up to .17/.18; or we need to wait until we can move everyone to 1.0.

For the record, I'm fine continuing to kick this can down the road, simply avoiding anything that locks us into *any* sdk version for the time being. That's easy to do: just rip the `opgenerate` target out of boilerplate and let the consuming repos define it. If we do that, we can still implement this PR. If a consumer wants to opt out of the osdk piece, they can just make that target a no-op.

2uasimojo:
Discussed in today's sync and agreed to support a range of osdk versions for the time being by discovering the desired version form go.mod and conditioning the `make opgenerate` rule accordingly. We will defer designing more perfect solutions for later.

bdematte:
As discussed on Slack, removing the support of v0.8.2. 
With regards to operator-sdk version support, efforts will rather focus on operator-sdk v1.0

boilerplate/openshift/golang-osd-operator/README.md
```diff
@@ -47,4 +47,14 @@ run code coverage analysis per [this SOP](https://github.com/openshift/ops-sop/b
 - runs it against
 - a `golangci.yml` config.
 
+## Checks on generated code
+
+The convention embeds default checks to ensure generated code generation is current, committed, and unaltered.
+To trigger the check, you can use `make check-generate` provided your Makefile properly includes the boilerplate-generated include `boilerplate/generated-includes.mk`.
```
2uasimojo, line 53: Bleh, `check-generate` isn't really noun-verb. However, I still have a problem with using the noun form of `generate` (`generation`) because I feel it's easily confused with the "resource version" definition. I don't know that `make generate-check` is really any better, since those are both verbs -- even though the protocol seems to be putting "the thing that's being done" -- in this case the `check` -- last.

I throw up my hands. Ignore the above.

"A foolish consistency is the hobgoblin of little minds, adored by little statesmen and philosophers and divines."
-- Emerson

boilerplate/openshift/golang-osd-operator/operator-sdk-generate.sh
```diff
@@ -20,10 +20,6 @@ VER=$(osdk_version $OSDK)
 # This explicitly lists the versions we know about. We don't support
 # anything outside of that.
 case $VER in
-  'v0.8.2')
-      $OSDK generate openapi
-      $OSDK generate k8s
-      ;;
```
2uasimojo, line 26: I don't think I agree with removing this support. Deprecating v0.8.2 and leaving it untested is one thing. Removing this entirely will slow adoption by gcp-project-operator. Can we leave it in until they've upgraded?

test/case/convention/openshift/golang-osd-operator/01_generated_files_checker
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash -e
+
+REPO_ROOT=$(git rev-parse --show-toplevel)
+
+source $REPO_ROOT/test/lib.sh
+
+for version in v0.15.1 v0.16.0 v0.17.0 v0.17.1 ; do
+    echo "Testing with operatore-sdk ${version}"
```
2uasimojo, line 8:
```suggestion
    echo "Testing with operator-sdk ${version}"
```

test/case/convention/openshift/golang-osd-operator/01_generated_files_checker
```diff
@@ -0,0 +1,35 @@
...
+    add_cleanup $repo
+    test_project="file-generate"
+
+    bootstrap_project $repo ${test_project} openshift/golang-osd-operator
+    cd $repo
+    sed s/__operator_sdk_version__/${version}/g go.mod.tmpl > go.mod
+    export GOPATH=$GOPATH:`pwd`
+
+    make update-boilerplate
+    make check-generate || echo "Project is not committed. Checks failed as expected"
```
2uasimojo, line 19: Ah, this one is supposed to fail on the `isclean` check. Can we maybe validate that part of the output?

test/case/convention/openshift/golang-osd-operator/01_generated_files_checker
```diff
@@ -0,0 +1,35 @@
...
+
+    git add -A
+    git commit -m "initial commit"
+
+    make check-generate || echo "Generated files are not committed. Checks failed as expected"
```
2uasimojo, line 24: ...And then this one passes `isclean` but fails since it does the generating anew.

test/lib.sh
```diff
@@ -48,6 +48,35 @@ empty_repo() {
...
+# - Copies in boilerplate/update from $REPO_ROOT
+# - Creates an empty boilerplate/update.cfg
+# It DOES run the update in order to allow using boilerplate/generated-includes.mk
+#
+# :param PATH: An existing directory that has been `git init`ed, like
+#       what you get when you run `empty_repo`.
+# :param TEST_PROJECT: the test_project (from test/projects)
+# :param INITIAL_CONVENTION: The convention(s) to be used for initializing the 
+#       project. If it contains several conventions, they need to be passed 
+#.      between " (eg "test_convention/foo test_convemtion/bar")
```
2uasimojo, line 63:
```suggestion
#.      between " (eg "test_convention/foo test_convention/bar")
```

test/projects/README.md
```diff
@@ -0,0 +1,4 @@
+This repository aims at containing all the projects needed to test proper behaviour of the conventions. This is mainly used from `test/convention` scenarios, the objective being to have a few simple projects ensuring smooth integration of new boilerplate in consumers.
+
+#### Projects specifics
+- `file-generate` is an Openshift operator based on `operator-sdk v0.16`
```
2uasimojo, line 4: This is no longer true.

test/projects/file-generate/README.md
```diff
@@ -0,0 +1,5 @@
+This is a dummy project used to test the `check-generate` make target for go operator (like in `openshift/golang-osd-operator`).
+
+#### Projects main points
+- This is an operator initialized with `operator-sdk v0.16` in order to have all the `make generate` commands working properly
```
2uasimojo, line 4: Not `v0.16`. Templated to allow substitution by the test case.

test/projects/file-generate/go.mod.tmpl
```diff
@@ -0,0 +1,42 @@
+module file-generate
+
+go 1.13
+
+require (
+	github.com/operator-framework/operator-sdk __operator_sdk_version__
```
2uasimojo, line 6: Maybe put a comment here explaining what's going on.

test/projects/file-generate/go.mod.tmpl
```diff
@@ -0,0 +1,42 @@
...
+	k8s.io/kube-scheduler => k8s.io/kube-scheduler v0.0.0-20191016114748-65049c67a58b
+	k8s.io/kubectl => k8s.io/kubectl v0.0.0-20191016120415-2ed914427d51
+	k8s.io/kubelet => k8s.io/kubelet v0.0.0-20191016114556-7841ed97f1b2
+	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.0.0-20191016115753-cf0698c3a16b
+	k8s.io/metrics => k8s.io/metrics v0.0.0-20191016113814-3b1a734dba6e
+	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.0.0-20191016112829-06bb3c9d77c9
+)
+
+replace github.com/docker/docker => github.com/moby/moby v0.7.3-0.20190826074503-38ab9da00309 // Required by Helm
+
```
2uasimojo, line 41: It would be nice if the test case could exercise the `replace` part of osdk version detection. That can be a followon, though.

test/projects/file-generate/deploy/crds/mygroup.com_testkinds_crd.yaml
```diff
@@ -0,0 +1,42 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
2uasimojo, line 1: It would be cool if `op-generate` had a real effect as well as `go-generate`. E.g. remove this file, and prove that it gets generated by your test.

test/case/convention/openshift/golang-osd-operator/01_generated_files_checker
```diff
@@ -0,0 +1,35 @@
...
+
+    if [ ! -f src/test/test-generation ] || [ ! -f src/test/subfolder/test-generation-subfolder ] ; then
```
2uasimojo, line 26: (See below) So here you could also make sure the CRD got generated.

test/case/convention/openshift/golang-osd-operator/01_generated_files_checker
```diff
@@ -0,0 +1,35 @@
...
+    echo "Testing with operatore-sdk ${version}"
+    repo=$(empty_repo)
+    add_cleanup $repo
+    test_project="file-generate"
+
+    bootstrap_project $repo ${test_project} openshift/golang-osd-operator
+    cd $repo
+    sed s/__operator_sdk_version__/${version}/g go.mod.tmpl > go.mod
+    export GOPATH=$GOPATH:`pwd`
+
```
2uasimojo, line 17: Should we make sure the generated files don't exist yet?

2uasimojo (commented):
The only blocker here is that I don't think we should remove 0.8.2 from op-generate until gcp-project-operator is migrated -- even if we omit it from testing.

2uasimojo:
Agreed in sync to rip out 0.8.2 support, as done in this PR.

And to address the remaining comments in a followon.

/lgtm
/approve

2uasimojo:
@bdematte @arjunrn it looks like this needs an update for the git version we've started to use from https://github.com/openshift/release/pull/12158.

2uasimojo:
/hold

...to stop wasting CI resources on bot-driven retests until this is fixed.

2uasimojo:
/lgtm
/approve

@bdematte is going to address nonblocking comments in a fup.

boilerplate/openshift/golang-osd-operator/README.md
2uasimojo: ✓ #59

test/case/convention/openshift/golang-osd-operator/01_generated_files_checker
2uasimojo: ✓ #59

test/lib.sh
2uasimojo: ✓ #59

test/projects/file-generate/go.mod.tmpl
2uasimojo: ✓ #59

test/projects/file-generate/deploy/crds/mygroup.com_testkinds_crd.yaml
2uasimojo: ✓ #59

test/case/convention/openshift/golang-osd-operator/01_generated_files_checker
2uasimojo: ✓ #59

test/case/convention/openshift/golang-osd-operator/01_generated_files_checker
2uasimojo: ✓ #59

test/projects/file-generate/README.md
2uasimojo: In the [fup](https://github.com/openshift/boilerplate/pull/59) you left this in, but added explanation below it. I guess you're saying you set this test project up initially using 0.16? Not really sure how that's relevant. But meh.

PR #42, opened 2020-09-09
Use a venv for python

2uasimojo:
...so we can install modules as non-root.


2uasimojo:
We're going to move the venv construction into ensure.sh

/hold

2uasimojo:
Tested. Works.

/hold cancel

boilerplate/_lib/ensure.sh
```diff
@@ -22,6 +22,14 @@ golangci-lint)
         curl -sfL "${DOWNLOAD_URL}" | tar -C "${GOPATH}/bin" -zx --strip-components=1 "golangci-lint-${GOLANGCI_LINT_VERSION}-${GOOS}-amd64/golangci-lint"
     fi
     ;;
+venv)
+    # Set up a python virtual environment
+    python3 -m venv .venv
```
jharrington22, line 27: Where is this being created? I thought you didn't have permissions in the directory the user was running?

2uasimojo: The pwd is the repo root

jharrington22: ah :+1:

2uasimojo: ...based on having run this via a `make` from that directory.

dustman9000, line 27: We should be ensuring the python3 is installed explicitly and obtain the absolute path to it before attempting the venv install.

2uasimojo: Agree. That should be covered (along with installing the libs) as part of https://issues.redhat.com/browse/OSD-5074 / https://issues.redhat.com/browse/OSD-5096 / etc.

PR #43, opened 2020-09-09
YAML Validation Fixes

dustman9000:
* Split YAML validation into 2 parts: olm-deploy vs. checking all yaml files
* Make olm-deploy check required for gotest


jharrington22:
@dustman9000 lets hold for #42 since @2uasimojo is going to move the install into `ensure.sh`

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -108,9 +108,16 @@ coverage:
...
 
 .PHONY: test
-test: gotest yaml-validate
+test: gotest validate-olm-deploy-yaml
 
-.PHONY: yaml-validate
-yaml-validate:
+.PHONY: install-pyyaml
+install-pyyaml:
 	python3 -mpip install pyyaml
```
2uasimojo, line 115: This doesn't work in CI because it tries to install into the real py3 dist, which is owned by root.

Adding `--user` is supposed to be the way around that, but that *also* doesn't work because the CI user's `$HOME` isn't set, so it tries to install in `/.local`.

I'm working on the PR to do set up the venv so we can do these things there.

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -108,9 +108,16 @@ coverage:
...
+
+.PHONY: validate-olm-deploy-yaml
+validate-olm-deploy-yaml: install-pyyaml
+	python3 boilerplate/openshift/golang-osd-operator/validate-yaml.py $(shell git ls-files | egrep '^(deploy)/' | egrep '.*\.ya?ml')
```
2uasimojo, line 119: Since we're being inclusive rather than exclusive, `ls-files` lets you use shell globs.

The quotes are necessary to avoid premature shell expansion.

The wildcarding *does* recurse into subdirectories, which is what we want.

```suggestion
	python3 boilerplate/openshift/golang-osd-operator/validate-yaml.py $(shell git ls-files 'deploy/*.yaml' 'deploy/*.yml')
```

boilerplate/openshift/golang-osd-operator/standard.mk
jharrington22: Yeap this is covered in #42

boilerplate/openshift/golang-osd-operator/standard.mk
jharrington22: Couldn't you then `$(shell git ls-files | egrep '^deploy.*\.ya?ml$')` ?

2uasimojo:
#42 is merged.

Since you're respinning, would you please squash locally? Otherwise the commit messages get glommed together and come out ugly.

PR #44, opened 2020-09-10
`operator-sdk generate` at the proper version 

2uasimojo:
Enhances ensure.sh to recognize the `operator-sdk` directive. This discovers the version of operator-sdk used by the consuming repository (from go.mod) and makes it available via a symlink at `./.operator-sdk/bin/operator-sdk`. If there's already one installed in your `$PATH`, and it's the right version, we just symlink to it. Otherwise we download the appropriate binary.

Adds a script to the `openshift/golang-osd-operator` convention called `operator-sdk-generate.sh` that uses the above, and then executes the `operator-sdk generate` commands appropriate to the version.

Swaps out the guts of `make opgenerate` to invoke the above.

### DIFF
--- a/boilerplate/_lib/ensure.sh
+++ b/boilerplate/_lib/ensure.sh
@@ -1,11 +1,14 @@
 #!/bin/bash
 set -euo pipefail
 
+source ${0%/*}/common.sh
+
 GOLANGCI_LINT_VERSION="1.30.0"
 DEPENDENCY=${1:-}
 GOOS=$(go env GOOS)
 
 case "${DEPENDENCY}" in
+
 golangci-lint)
     GOPATH=$(go env GOPATH)
     if which golangci-lint ; then
@@ -22,6 +25,54 @@ golangci-lint)
         curl -sfL "${DOWNLOAD_URL}" | tar -C "${GOPATH}/bin" -zx --strip-components=1 "golangci-lint-${GOLANGCI_LINT_VERSION}-${GOOS}-amd64/golangci-lint"
     fi
     ;;
+
+operator-sdk)
+    #########################################################
+    # Ensure operator-sdk is installed at the desired version
+    # When done, ./.operator-sdk/bin/operator-sdk will be a
+    # symlink to the appropriate executable.
+    #########################################################
+    # First discover the desired version from go.mod
+    # The following properly takes `replace` directives into account.
+    wantver=$(go list -json -m github.com/operator-framework/operator-sdk | jq -r 'if .Replace != null then .Replace.Version else .Version end')
+    echo "go.mod says you want operator-sdk $wantver"
+    # Where we'll put our (binary and) symlink
+    mkdir -p .operator-sdk/bin
+    cd .operator-sdk/bin
+    # Discover existing, giving preference to one already installed in
+    # this path, because that has a higher probability of being the
+    # right one.
+    if [[ -x ./operator-sdk ]] && [[ "$(osdk_version ./operator-sdk)" == "$wantver" ]]; then
+        echo "operator-sdk $wantver already installed"
+        exit 0
+    fi
+    # Is there one in $PATH?
+    if which operator-sdk && [[ $(osdk_version $(which operator-sdk)) == "$wantver" ]]; then
+        osdk=$(realpath $(which operator-sdk))
+        echo "Found at $osdk"
+    else
+        case "$(uname -s)" in
+            Linux*)
+                binary="operator-sdk-${wantver}-x86_64-linux-gnu"
+                ;;
+            Darwin*)
+                binary="operator-sdk-${wantver}-x86_64-apple-darwin"
+                ;;
+            *)
+                echo "OS unsupported"
+                exit 1
+                ;;
+        esac
+        echo "Downloading $binary"
+        curl -OJL https://github.com/operator-framework/operator-sdk/releases/download/${wantver}/${binary}
+        chmod +x ${binary}
+        osdk=${binary}
+    fi
+    # Create (or overwrite) the symlink to the binary we discovered or
+    # downloaded above.
+    ln -sf $osdk operator-sdk
+    ;;
+
 venv)
     # Set up a python virtual environment
     python3 -m venv .venv
@@ -30,6 +81,7 @@ venv)
         .venv/bin/python3 -m pip install -r "$2"
     fi
     ;;
+
 *)
     echo "Unknown dependency: ${DEPENDENCY}"
     exit 1
--- a/boilerplate/openshift/golang-osd-operator/operator-sdk-generate.sh
+++ b/boilerplate/openshift/golang-osd-operator/operator-sdk-generate.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+set -euo pipefail
+
+###
+# Run operator-sdk generate commands appropriate to the version of
+# operator-sdk configured in the consuming repository.
+###
+
+REPO_ROOT=$(git rev-parse --show-toplevel)
+LIB=$REPO_ROOT/boilerplate/_lib
+
+source $LIB/common.sh
+
+$LIB/ensure.sh operator-sdk
+
+# Symlink to operator-sdk binary set up by `ensure.sh operator-sdk`:
+OSDK=$REPO_ROOT/.operator-sdk/bin/operator-sdk
+
+VER=$(osdk_version $OSDK)
+
+# This explicitly lists the versions we know about. We don't support
+# anything outside of that.
+case $VER in
+  'v0.8.2')
+      $OSDK generate openapi
+      $OSDK generate k8s
+      ;;
+  'v0.15.1'|'v0.16.0'|'v0.17.0'|'v0.17.1')
+      $OSDK generate crds
+      $OSDK generate k8s
+      ;;
+  *) err "Unsupported operator-sdk version $VER" ;;
+esac
--- a/boilerplate/openshift/golang-osd-operator/standard.mk
+++ b/boilerplate/openshift/golang-osd-operator/standard.mk
@@ -91,8 +91,7 @@ go-generate:
 
 .PHONY: op-generate
 op-generate:
-	operator-sdk generate crds
-	operator-sdk generate k8s
+	${CONVENTION_DIR}/operator-sdk-generate.sh
 	# Don't forget to commit generated files
 
 .PHONY: generate
--- a/boilerplate/_lib/common.sh
+++ b/boilerplate/_lib/common.sh
@@ -3,6 +3,19 @@ err() {
   exit 1
 }
 
+## osdk_version BINARY
+#
+# Print the version of the specified operator-sdk BINARY
+osdk_version() {
+    local osdk=$1
+    # `operator-sdk version` output looks like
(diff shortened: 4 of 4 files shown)


bdematte:
For now (at least until `operator-sdk 1.0` migration), we can rely on the project to determine the version of the sdk because of commands removal/replacement. 
I just tested to run the `01_generated_files_checker` with operator-sdk 0.19 and it miserably fails because of commands which have been rename/changed compared to 0.16 .

boilerplate/_lib/ensure.sh
```diff
@@ -22,6 +36,54 @@ golangci-lint)
...
+    # Where we'll put our (binary and) symlink
+    mkdir -p .operator-sdk/bin
+    cd .operator-sdk/bin
+    # Discover existing, giving preference to one already installed in
+    # this path, because that has a higher probability of being the
+    # right one.
+    if [[ -x ./operator-sdk ]] && [[ "$(osdk_version ./operator-sdk)" == "$wantver" ]]; then
+        echo "operator-sdk $wantver already installed"
+        exit 0
+    fi
```
bdematte, line 59: In case operator-sdk version mismatches, it may be good to error (with a variable to skip the test?) to avoid confusion as the user running the `operator sdk` manually would lead to a different behaviour).

boilerplate/_lib/ensure.sh
```diff
@@ -22,6 +36,54 @@ golangci-lint)
...
+
+operator-sdk)
+    #########################################################
+    # Ensure operator-sdk is installed at the desired version
+    # When done, ./.operator-sdk/bin/operator-sdk will be a
+    # symlink to the appropriate executable.
+    #########################################################
+    # First discover the desired version from go.mod
+    # The following properly takes `replace` directives into account.
+    wantver=$(go list -json -m github.com/operator-framework/operator-sdk | jq -r 'if .Replace != null then .Replace.Version else .Version end')
```
jharrington22, line 48: A couple of things, this will list a version for the operator-sdk is an indirect dependency, we would want to ensure that `.Indirect != true` 

If operator-sdk is not a dep then this command will error with "go list -m: module github.com/operator-framework/operator-sdk: not a known dependency" though `wantevr` is handled if its null should we hide the error output and let the user know that operator-sdk is not a required dep?

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -90,9 +90,16 @@ gogenerate:
...
+opgenerate: operator-sdk
+	# NOTE: operator-sdk generate happily prints a usage statement
+	# and exits zero when given an unknown target. Running generate
+	# commands appropriate to all known versions of operator-sdk.
+	# This ought to work for everyone, though it might print ugly
+	# output for some. We'll be able to clean this up when we
+	# standardize on the operator-sdk version.
+	${OPERATOR_SDK} generate openapi
+	${OPERATOR_SDK} generate crds
+	${OPERATOR_SDK} generate k8s
```
jharrington22, line 102: Should we be error checking here?

boilerplate/_lib/ensure.sh
bdematte: mehh, don't fully agree. make targets should be a convenient way to run standard commands, but not the only easy one. Developers needs to be able to continue using standard tools easily.
If I need to debug and want to add a parameter to the `operator-sdk` command, I don't want to have to clone boilerplate, update it and then set the ENV variable overiding the GIT repo to finally succeed to run the `make` command. I should be able to easily run the same command (and here, version mismatch would be a pain for the user)

2uasimojo: Oh, absolutely. But if the developer is running their own binary, they're responsible for making sure it's the right version. There's nothing we can do to help them with that -- except for providing a convenient `make` target.

2uasimojo:
Per [discussion](https://github.com/openshift/boilerplate/pull/41#issuecomment-692119739) on #41, I'm working on updating this to accommodate a range of versions.

I went through all of the [repositories](https://github.com/openshift/boilerplate/blob/2a5017b3b52b92969ab4799c90df8e3fa39557a9/subscribers.yaml) I think we're going to be dealing with. They're currently using five different versions of operator-sdk, and those five versions support two different permutations of `generate`:

| versions | generators |
|-|-|
| 0.8.2 | k8s, openapi |
| 0.15.1, 0.16.0, 0.17.0, 0.17.1 | crds, k8s |

I'll make it so.

2uasimojo:
Tested against repos at each of the five supported versions.
Also tested error cases:
- Not a modules-based project
- Version that exists, but that we don't support
- Bogus version

This is ready to go.

/hold cancel

boilerplate/openshift/golang-osd-operator/operator-sdk-generate.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
dustman9000, line 1: I'd propose putting it under `_lib` as this is essentially acting as a common library function now. It will also allow it to be used by future `golang` `operator-sdk` type conventions.

2uasimojo: It'll apply to golang operators, so would stay with all the other stuff in the existing convention when we do whatever to split those out.

But we can move it later as needed.

Done.

bdematte: We main need to allign on the purpose of _lib. The way I see it, it contains containing what I would call 'technical pieces' like the ensure.sh script (used by all conventions and fulfilling a generic mission) 
I don't like starting to move functional pieces in _lib, it will make the maintenance more complex as no clue on impacted convention(s) and make the purpose of _lib more obscure. 
At least, we should clarify what we expect in _lib and what we don't expect in it to be alligned.

2uasimojo: I agree, @bdematte. We've talked about reorganizing the directory structure of the main `boilerplate/` subdirectory to avoid clutter; but IMO this script clearly belongs inside the convention regardless.

Anyone else want to render an opinion? Else I'll move it back.

dustman9000: I don't think his question was really answered. What is the criteria for putting a script in the `_lib` directory vs. the convention directories?

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -91,8 +91,7 @@ gogenerate:
 
 .PHONY: opgenerate
```
dustman9000, line 92: I'm just now seeing these Make targets. We seem to be very inconsistent in our naming convention here. Can we standardize on `<noun>-<verb>` in kebab-case? This would be consistent with `docker-build` and other targets. It would also make it more readable.
This would become `op-generate`...

2uasimojo: +1. Do you want to iterate on those names here or in the [card](https://issues.redhat.com/browse/OSD-4673)?

boilerplate/openshift/golang-osd-operator/operator-sdk-generate.sh
2uasimojo: My take: `_lib` is for things relevant to the boilerplate infrastructure; convention is for things relevant to the consumer itself. There will be some overlap there; IMO convention takes precedence.

IOW: If it ever gets run by the consuming repository *other than* when they're explicitly updating boilerplate, it should go in a convention directory.

In this case, the consumer can run `make opgenerate` any time, not just as part of the "freeze check" we're implementing via #41.

(By this definition, `ensure.sh` is still a bit nebulous. The things it's ensuring right now are in the path of convention-owned `make` targets of general usefulness by the consumer. So we could make a case that it should go in the convention directory. We can do that in another PR if there's consensus on that. @arjunrn wdyt?)

boilerplate/openshift/golang-osd-operator/standard.mk
2uasimojo: Sorry, I just twigged that changing these target names isn't relevant to this PR at all, and only partly overlapping with the card referenced above (and its PR, #49).

Still +1 on renaming targets to follow some kind of pattern. But it needs to be done in a separate PR so we can keep this one focussed. Please do comment on [OSD-4673](https://issues.redhat.com/browse/OSD-4673) or #49 to refine the names of the targets to be used by prow and appsre; and propose any other renames separately.

boilerplate/_lib/ensure.sh
```diff
@@ -1,11 +1,14 @@
 #!/bin/bash
```
dustman9000, line 1: IMO, this entire `ensure.sh` should really just go away. This is really what Makefile dependency chains are for. Why can't we run these commands under Makefile targets and have a proper chain?

For example:
```
operator-sdk-bin: /path/to/operator-sdk
    #install operator-sdk commands

opgenerate: operator-sdk-bin
    $(OPERATOR_SDK) generate $(OPERATOR_SDK_TARGETS)
```

boilerplate/openshift/golang-osd-operator/standard.mk
dustman9000: Fair enough, I commented on #49 to hopefully clean it up there. Thanks.

boilerplate/_lib/ensure.sh
2uasimojo: I did play with something [close to that](https://github.com/openshift/boilerplate/commit/689cbda3fc7649ea06e77d7ef343061de2bdabad#diff-6efb8dd653c9062efcfb26adab7756fc) at one point in this PR, and I'm okay going back in that direction if folks feel strongly about it. But it gets nowhere near eliminating `ensure.sh`, and IMO it shouldn't.

`make` simply isn't designed for complex branching/looping. Yes, it has conditionals, but they're awkward to use and read, especially when they need to be nested *. And while `ensure.sh` isn't super complicated, each of its cases has enough logic to be awkward if ported to native `make`.

\* Using bash conditionals in the target doesn't count. I don't object to a target that runs multiple commands, but I'm a hard no on "one-line shell scripts" embedded in `make`. For example, [this](https://github.com/openshift/must-gather-operator/blob/49dd9f1fa3321b9506478fcd56970635ecb871cb/functions.mk) has got to go.

boilerplate/openshift/golang-osd-operator/operator-sdk-generate.sh
arjunrn: @2uasimojo I agree, I only realized later that `_lib` was not the right place for that script.

2uasimojo: > @2uasimojo I agree, I only realized later that `_lib` was not the right place for that script.

Cool. Moved via #52.

So are we all okay with operator-sdk-generate.sh living in the convention dir as well?

2uasimojo: Reverted.

2uasimojo:
/test pr-check

One more time and I'm gonna sic dptp on ya.

2uasimojo:
Let's see if the rebase kicks the tests happy.

PR #45, opened 2020-09-11
Added dockerfile with dependencies

arjunrn:
The docker image would contain the following dependencies:

1. golangci-lint
2. operator-sdk CLI
3. opm
4. PyYAML for python3
5. git v2.28.0

Once the Docker is merged into master it can be built and published by PROW. Once published we can use specify that image in the `docker run`/`podman run` parts. This will be done in subsequent PRs.

### DIFF
--- a/config/build.sh
+++ b/config/build.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+set -euo pipefail
+
+GOCILINT_VERSION="1.31.0"
+GOCILINT_SHA256SUM="9a5d47b51442d68b718af4c7350f4406cdc087e2236a5b9ae52f37aebede6cb3"
+GOCILINT_LOCATION=https://github.com/golangci/golangci-lint/releases/download/v${GOCILINT_VERSION}/golangci-lint-${GOCILINT_VERSION}-linux-amd64.tar.gz
+
+OPERATOR_SDK_VERSION="0.16.0"
+OPERATOR_SDK_SHA256SUM="3df782f341749f7962ab0fcfedd2961c18b21ad34ff7acd194b49a152f59abcb"
+OPERATOR_SDK_LOCATION=https://github.com/operator-framework/operator-sdk/releases/download/v${OPERATOR_SDK_VERSION}/operator-sdk-v${OPERATOR_SDK_VERSION}-x86_64-linux-gnu
+
+OPM_VERSION="1.13.8"
+OPM_SHASUM="a48aa9d69b0be3439220e818edde4b36b3b9eceb2a058d86b4fdf0ca9dcd21c8"
+OPM_LOCATION=https://github.com/operator-framework/operator-registry/releases/download/v${OPM_VERSION}/linux-amd64-opm
+
+curl -L -o golangci-lint.tar.gz $GOCILINT_LOCATION
+echo ${GOCILINT_SHA256SUM} golangci-lint.tar.gz | sha256sum -c
+tar xzf golangci-lint.tar.gz golangci-lint-${GOCILINT_VERSION}-linux-amd64/golangci-lint
+mv golangci-lint-${GOCILINT_VERSION}-linux-amd64/golangci-lint /usr/local/bin
+rm -rf golangci-lint-${GOCILINT_VERSION}-linux-amd64
+rm -f golangci-lint.tar.gz
+
+curl -L -o operator-sdk $OPERATOR_SDK_LOCATION
+echo ${OPERATOR_SDK_SHA256SUM} operator-sdk | sha256sum -c
+mv operator-sdk /usr/local/bin
+
+curl -L -o opm $OPM_LOCATION && \
+echo ${OPM_SHASUM} opm | sha256sum -c
+mv opm /usr/local/bin
+
+python3 -m pip install PyYAML==5.3.1
+
+GIT_VERSION="2.28.0"
+GIT_SHASUM="02016d16dbce553699db5c9c04f6d13a3f50727c652061b7eb97a828d045e534"
+GIT_DEPENDENCIES="epel-release perl-CPAN gettext-devel perl-devel openssl-devel zlib-devel curl-devel expat-devel getopt asciidoc xmlto docbook2X"
+yum remove -y git*
+yum -y install ${GIT_DEPENDENCIES}
+yum -y groupinstall "Development Tools"
+curl -L -o git.tar.gz https://github.com/git/git/archive/v${GIT_VERSION}.tar.gz
+echo "${GIT_SHASUM}" git.tar.gz | sha256sum -c
+tar xzf git.tar.gz
+make --directory "git-${GIT_VERSION}" configure
+./git-${GIT_VERSION}/configure --prefix=/usr
+make --directory "git-${GIT_VERSION}"
+make --directory "git-${GIT_VERSION}" install 
+rm -rf "git-${GIT_VERSION}"
+yum groupremove -y "Development Tools" && \
+yum -y remove ${GIT_DEPENDENCIES}
+yum clean all
+yum -y autoremove
+rm -rf /var/cache/yum
--- a/config/Dockerfile
+++ b/config/Dockerfile
@@ -1,0 +1,4 @@
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.13
+
+COPY build.sh /build.sh
+RUN /build.sh && rm -f /build.sh


boilerplate/_lib/Dockerfile
```diff
@@ -0,0 +1,40 @@
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.13
+
+ENV GOCILINT_VERSION="1.31.0"
```
jharrington22, line 3: I like that specific versions of all these packages are supplied. Should we move them out into a config file imported by make and supply these as build args to docker? Should boilerplate have a common config file with versions of all dependencies it uses?

boilerplate/_lib/Dockerfile
```diff
@@ -0,0 +1,40 @@
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.13
```
jharrington22, line 1: We use this as a base to build go and then use `registry.access.redhat.com/ubi8/ubi-minimal:latest` on top for operators and CI containers. Can we do that here it may also help with packages below that I assume are missing from yum?

boilerplate/update
```diff
@@ -37,7 +37,7 @@ else
   echo "Updating the update script."
   rsync -a "${OUT}/boilerplate/update" "$0"
   echo "Copying utilities"
-  rsync -a -r ${OUT}/boilerplate/_* $DIR
+  rsync -a -r --delete ${OUT}/boilerplate/_* $DIR
```
jharrington22, line 40: Probably should be a separate PR.

arjunrn: I've made #46 and fixup this PR once that is merged.

boilerplate/_lib/Dockerfile
arjunrn: It looks like prow [does not support](https://github.com/openshift/ci-tools/blob/master/CONFIGURATION.md#images) build arguments for docker images at the moment. So these versions will have to stay here for the time being.

jharrington22: Just concerned about having multiple sources of truth for dependency versions.

boilerplate/update
jharrington22: Thanks!

boilerplate/_lib/Dockerfile
arjunrn: Once we containerize all the make targets we can remove all the other scripts/helpers which install dependencies and then the Dockerfile should be the source-of-truth.

boilerplate/_lib/Dockerfile
arjunrn: It is possible but I don't think it's necessary. Also it will complicate stuff because we will have to copy the git binaries from the first to second stage. So stuff like shell autocompletion may not work. Also it's probably better to have a full featured image to run the boilerplate stuff esp. for stuff like debugging. And since this will never be run in production it doesn't pose a security risk. The only problem is size of the image.

2uasimojo: No objection here, but "shell autocompletion" caught my eye. Under what circumstances is that relevant? This image is for running containerized automated builds & checks, and for running the operator, right? Should be non-interactive except for the 0.01% of the time we need to jump in and debug something?

boilerplate/_lib/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+    echo ${GOCILINT_SHA256SUM} golangci-lint.tar.gz && \
+    tar xzf golangci-lint.tar.gz golangci-lint-${GOCILINT_VERSION}-linux-amd64/golangci-lint && \
+    mv golangci-lint-${GOCILINT_VERSION}-linux-amd64/golangci-lint /usr/local/bin && rm -rf golangci-lint-${GOCILINT_VERSION}-linux-amd64 && \
+    rm -f golangci-lint.tar.gz && \
+    curl -L -o operator-sdk https://github.com/operator-framework/operator-sdk/releases/download/v${OPERATOR_SDK_VERSION}/operator-sdk-v${OPERATOR_SDK_VERSION}-x86_64-linux-gnu && \
+    echo ${OPERATOR_SDK_SHA256SUM} operator-sdk | sha256sum -c && \
+    mv operator-sdk /usr/local/bin && \
+    curl -L -o opm https://github.com/operator-framework/operator-registry/releases/download/v${OPM_VERSION}/linux-amd64-opm && \
+    echo ${OPM_SHASUM} opm | sha256sum -c && \
+    mv opm /usr/local/bin
```
2uasimojo, line 22: Can we move these one-massive-line commands into a script in the repo next to this Dockerfile? Or is that a chicken/egg problem?

2uasimojo (commented):
I like where this is going, for sure.

Can we talk about the location of this Dockerfile? I'm not sure `_lib` makes sense; IMO it should go next to things like the `update` script. But perhaps this is a good time to crisp up the directory structure as we've discussed previously.

Also: this PR doesn't contain anything that *uses* this Dockerfile. Is that going to live in prow? (How) will it be used by bp consumers eventually? Can we call those details out in the commit message please?

Also: Please squash locally. Otherwise we'll have an ugly three-part commit message with irrelevant gorp in it.

boilerplate/_lib/Dockerfile
arjunrn: It can be done.

boilerplate/_lib/Dockerfile
arjunrn: Shell autocompletion is not the only thing. I'm no C++ expert but the way I understand it when a C++ binary is built it is linked to libraries on the machine. Simply copying the binary may not work like in the case of Go.

config/build.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+set -euo pipefail
+
+GOCILINT_VERSION="1.31.0"
+GOCILINT_SHA256SUM="9a5d47b51442d68b718af4c7350f4406cdc087e2236a5b9ae52f37aebede6cb3"
+GOCILINT_LOCATION=https://github.com/golangci/golangci-lint/releases/download/v${GOCILINT_VERSION}/golangci-lint-${GOCILINT_VERSION}-linux-amd64.tar.gz
+
+OPERATOR_SDK_VERSION="0.16.0"
```
2uasimojo, line 8: This will make the image unusable for `test` from consumers using a different version once #41 is in place. (For building it's irrelevant, so fine.)

config/build.sh
```diff
@@ -0,0 +1,51 @@
...
+curl -L -o golangci-lint.tar.gz $GOCILINT_LOCATION
+echo ${GOCILINT_SHA256SUM} golangci-lint.tar.gz | sha256sum -c
+tar xzf golangci-lint.tar.gz golangci-lint-${GOCILINT_VERSION}-linux-amd64/golangci-lint
+mv golangci-lint-${GOCILINT_VERSION}-linux-amd64/golangci-lint /usr/local/bin
+rm -rf golangci-lint-${GOCILINT_VERSION}-linux-amd64
+rm -f golangci-lint.tar.gz
+
+curl -L -o operator-sdk $OPERATOR_SDK_LOCATION
+echo ${OPERATOR_SDK_SHA256SUM} operator-sdk | sha256sum -c
+mv operator-sdk /usr/local/bin
```
2uasimojo, line 25: We just got bitten by this, so I want to make sure: are we root at this point? Or is `/usr/local/bin` otherwise somehow writable?

Also, are we sure `/usr/local/bin` appears early enough in `$PATH` that we'll get the version we're installing even if there's another one installed already? And/or are we sure there's not another one installed already? (Probably the latter, or we wouldn't be here ;P )

2uasimojo (approved):
Thanks @arjunrn. I didn't try to run this or vet that it has everything we need, but it's certainly not going to hurt anything yet, so let's land it as at least a starting point.

/lgtm
/approve

config/build.sh
arjunrn: Yes, the docker build is done as root, so everything in writable. You make a good point about preexisting executables. I can add an extra step which delete executable on the path with the same name.

dustman9000:
Why are we putting a Dockerfile in this repo? Are we really planning to build and deploy artifacts from this repo?

arjunrn:
> Why are we putting a Dockerfile in this repo? Are we really planning to build and deploy artifacts from this repo?

@dustman9000 We need a docker image not only for testing but also published so that users of boilerplate can run the make targets in a container with all dependencies. The image added will be used for the later. For testing we can reuse the same image but that will be in a separate PR.

PR #49, opened 2020-09-14
Organize targets for prow and appsre

2uasimojo:
Per the referenced Jira, we want to standardize the prow configurations for all repositories consuming openshift/golang-osd-operator. This commit refactors the targets to have the desired names and (mostly) do the right things for both prow and appsre. (Since this is primarily a refactor, it leaves TODOs for several things requiring heavy lifting.)

Jira: [OSD-4673](https://issues.redhat.com/browse/OSD-4673)

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/standard.mk
+++ b/boilerplate/openshift/golang-osd-operator/standard.mk
@@ -63,24 +63,21 @@ clean:
 isclean:
 	@(test "$(ALLOW_DIRTY_CHECKOUT)" != "false" || test 0 -eq $$(git status --porcelain | wc -l)) || (echo "Local git checkout is not clean, commit changes and try again." >&2 && exit 1)
 
-.PHONY: build
-build: isclean
+.PHONY: docker-build
+docker-build: isclean
 	${CONTAINER_ENGINE} build . -f $(OPERATOR_DOCKERFILE) -t $(OPERATOR_IMAGE_URI)
 	${CONTAINER_ENGINE} tag $(OPERATOR_IMAGE_URI) $(OPERATOR_IMAGE_URI_LATEST)
 
-.PHONY: push
-push:
+.PHONY: docker-push
+docker-push:
 	${CONTAINER_ENGINE} push $(OPERATOR_IMAGE_URI)
 	${CONTAINER_ENGINE} push $(OPERATOR_IMAGE_URI_LATEST)
 
-# These names are used by the app-sre pipeline targets
-.PHONY: docker-build
-docker-build: build
-.PHONY: docker-push
-docker-push: push
+.PHONY: push
+push: docker-push
 
 .PHONY: go-check
-go-check: ## Lint code
+go-check: ## Golang linting and other static analysis
 	${CONVENTION_DIR}/ensure.sh golangci-lint
 	GOLANGCI_LINT_CACHE=${GOLANGCI_LINT_CACHE} golangci-lint run -c ${CONVENTION_DIR}/golangci.yml ./...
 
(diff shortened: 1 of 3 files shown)


boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -125,3 +115,44 @@ yaml-validate: python-venv
...
 validate-olm-deploy-yaml: python-venv
 	${PYTHON} ${CONVENTION_DIR}/validate-yaml.py $(shell git ls-files 'deploy/*.yaml' 'deploy/*.yml')
+
+######################
+# Targets used by prow
+######################
+
+# validate: Ensure code generation has not been forgotten; and ensure
+# generated and boilerplate code has not been modified.
+# TODO
```
dustman9000, line 125: Missing `TODO` content?

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -125,3 +115,44 @@ yaml-validate: python-venv
...
+#########################
+# Targets used by app-sre
+#########################
+
+# build-push: Construct, tag, and push the official operator and
+# registry container images.
+# TODO: Boilerplate this script.
+.PHONY: build-push
+build-push:
+	hack/app_sre_build_deploy.sh
```
dustman9000, line 158: This needs to be the path under `${CONVENTION_DIR}` not the consumer repo's hack directory.

boilerplate/openshift/golang-osd-operator/standard.mk
2uasimojo: This is saying the guts of the whole target still need to be done.

boilerplate/openshift/golang-osd-operator/standard.mk
2uasimojo: Not yet. A separate PR will boilerplate the script (per the TODO) -- but it'll have to do it in such a way that it _still_ calls out to the repository's bespoke bundle generation. And then a PR after _that_ will figure out how to do that bundle gen generically from within boilerplate.

All this is trying to accomplish right now is allow us to change lines like [this one](https://gitlab.cee.redhat.com/service/app-interface/-/blob/030a6acfc5c614796c7a64010ef34d75a9365f5f/data/services/osd-operators/cicd/ci-int/jobs-aws-account-operator.yaml#L24) to say `make build-push` instead. Then we'll be able to swap out what's underneath that without having to keep changing app-interface.

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -63,24 +63,21 @@ clean:
...
 
-# These names are used by the app-sre pipeline targets
-.PHONY: docker-build
-docker-build: build
-.PHONY: docker-push
-docker-push: push
+.PHONY: push
+push: docker-push
 
 .PHONY: gocheck
```
dustman9000, line 79: Can we standardize on a naming convention here. I'd prefer something like `<noun>-<verb>` in kebab-case for consistency. 
Can we change all of the follow to `go-<verb>` please?

2uasimojo: Done via #51. If you have suggestions for how to rename the targets added in this PR to be `noun-verb`ish, I'll update this one.

boilerplate/openshift/golang-osd-operator/standard.mk
2uasimojo: I added more words here, since I was rebasing.

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -124,3 +114,46 @@ yaml-validate: python-venv
 .PHONY: olm-deploy-yaml-validate
 olm-deploy-yaml-validate: python-venv
 	${PYTHON} ${CONVENTION_DIR}/validate-yaml.py $(shell git ls-files 'deploy/*.yaml' 'deploy/*.yml')
+
+######################
```
dustman9000, line 118: Should we separate these into a `prow.mk` to reduce file bloat?

2uasimojo: I actually talked about that with @jharrington22 before pulling the trigger on this PR. He preferred to keep it monolithic for searchability. I'd be happy to go either way.

With the nexus makefile include, it'd be trivial to move later if we feel like it.

PR #50, opened 2020-09-15
`make clone-this`

2uasimojo:
To facilitate testing local changes, you can now run

```shell
$ eval `make clone-this`
```

and it will set the `BOILERPLATE_GIT_REPO` environment variable to point to your local boilerplate clone.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,3 +14,11 @@ test: isclean
 
 .PHONY: pr-check
 pr-check: test
+
+# Use as:
+#     eval $(make clone-this)
+# Then you can go to a consuming repo (in the same shell) and run
+# `make update-boilerplate` to test your local changes.
+.PHONY: clone-this
+clone-this:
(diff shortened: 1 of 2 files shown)


bdematte:
I think this is a bit overkill. It is not making it fewer commands but adds an extra indirection reducing readability.
If we want a global target to set a bunch of inter-connected variables, we can. But 1 target per env variable, I am not convinced.

2uasimojo:
You’re right.  When I started this, it was going to generate the `BOILERPLATE_GIT_CLONE` variable instead, set up to grab the current branch of your fork from the server. That was worth factoring out, as it involved some nontrivial commands and flags. But I decided that didn’t have significant value for iterative development versus using your local directory, so I ended up here. The only value is me not having to look up how the env var is spelled (I can use autocompletion for the target) — but that’s not worth it.

PR #51, opened 2020-09-16
Make `make` targets $noun-$verb where reasonable

2uasimojo:
In the spirit of e.g. `docker-build`, rename `make` targets to conform to a `$noun-$verb` format where it makes sense.


dustman9000:
/lgtm
Need to communicate target changes to consuming repos and updated [openshift/release](https://github.com/openshift/release)

PR #52, opened 2020-09-16
Move ensure.sh into openshift/golang-osd-operator

2uasimojo:
This script is used by `make` targets in `openshift/golang-osd-operator`, not by boilerplate infrastructure, so it makes more sense for it to live in the convention.

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/ensure.sh
+++ b/boilerplate/openshift/golang-osd-operator/ensure.sh
@@ -1,7 +1,8 @@
 #!/bin/bash
 set -euo pipefail
 
-source ${0%/*}/common.sh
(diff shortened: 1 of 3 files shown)


2uasimojo:
/hold

We've reached consensus on #44, so let's land that one first. This PR will then need to be updated to also change the [reference in operator-sdk-generate.sh](https://github.com/openshift/boilerplate/pull/44/files#diff-0abaddc254dbab5ac2300b00683eb299R14).

2uasimojo:
Well, that rebase went horribly wrong.

2uasimojo:
Phew, that was harder than it should have been. But it's sorted now.

/hold cancel

PR #53, opened 2020-09-17
`update` creates a commit

2uasimojo:
The main `update` driver is modified to create a git commit containing the changes it brings in.

Because we're committing, it becomes more important that the consuming checkout is clean before we start, so that it's easy to revert if something goes wrong. A hard check is added for that. However, we make an exception for update.cfg so that you don't have to have a whole commit with just a delta to that file.

In the spirit of the famed and fabled `promote.sh`, the generated commit's message contains a history of the boilerplate commits since the last update. But it also handles an initial bootstrap sanely (you don't get the full history of the boilerplate repo). It also lists the conventions processed.

A granny switch is provided to avoid the automated commit:
`export BOILERPLATE_UPDATE_NO_COMMIT=1`


dustman9000:
This is a good idea, but I really don't understand why we need to perform a git commit on behalf of the user like this. This could be potentially destructive to a local workspace and assumes quite a few things on behalf of the user. I'd prefer to keep the consumer's adoption of boilerplate explicit to the user if we can. This has the potential to be too heavy handed and prevent adoption.

2uasimojo:
Discussing in [slack](https://coreos.slack.com/archives/C010DRY1ZQ9/p1600380385307800). I'll summarize here when we reach consensus.

2uasimojo:
Consensus is that this is a good thing, but we need the ability to run the commit piece separately.

That's turning out to be different enough from what's in here that I'm working it in a separate PR: #56.

PR #58, opened 2020-09-28
Configure golangci-lint to keep going

2uasimojo:
By default, golangci-lint will stop printing errors after a certain limit. This can be annoying, especially in CI, as you would have to keep running it repeatedly to fix all the problems. So this commit adds configuration values to disable those limits, meaning *all* errors will be printed on each run.


arjunrn:
:+1:

PR #61, opened 2020-10-01
Don't break lint if there is no yaml to validate.

jewzaam:
Make targets pass in files to validate but not all repos have yaml that needs validating.
In that case the yaml validation fails because it prints usage and exits with 1.

This PR simply changes the exit code to 0 so it's not going to fail.

There is no great alternative to this, we don't know where yaml might be or if there might be a directory to pass in etc.  Proposing we simply don't fail without arguments.

blocks https://github.com/openshift/configure-alertmanager-operator/pull/101


2uasimojo:
Thanks @jewzaam 

/lgtm
/approve

PR #63, opened 2020-10-05
Make `make validate` include `freeze-check`

2uasimojo:
...to ensure boilerplate-owned files are not changed.

This adds a `make` target in the boilerplate framework itself called `boilerplate-freeze-check` and includes it in the chain of the `validate` target in the `openshift/golang-osd-operator` convention.

Jira: [OSD-5366](https://issues.redhat.com/browse/OSD-5366)

### DIFF
--- a/test/lib.sh
+++ b/test/lib.sh
@@ -295,3 +295,31 @@ last_commit_message() {
         git log -1 | tail -n +5
     )
 }
+
+## expect_failure ERROR_STRING CMD ARGS...
+#
+# Runs CMD with ARGS...
+# Fails if CMD succeeds.
+# If CMD fails, the output (stdout+stderr) is grepped for ERROR_STRING,
+# and we succeed iff it is found.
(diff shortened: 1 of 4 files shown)


test/lib.sh
```diff
@@ -295,3 +295,31 @@ last_commit_message() {
...
+
+## expect_failure ERROR_STRING CMD ARGS...
+#
+# Runs CMD with ARGS...
+# Fails if CMD succeeds.
+# If CMD fails, the output (stdout+stderr) is grepped for ERROR_STRING,
+# and we succeed iff it is found.
+# CMD ARGS... is run via eval "$@". Quote accordingly.
+expect_failure () {
+  local errstr=$1
```
dustman9000, line 307: nit: seems we are at 4 spaces here for indentation

2uasimojo: Ack, will fix. I should just set my vim default...

2uasimojo: Done.

PR #68, opened 2020-10-09
Updates to standard.mk to support building on other platforms and

rrati:
broke down operator IMG into usable sub-components


2uasimojo (approved):
Beautiful.

This is as discussed in https://github.com/openshift/release/pull/12645

/lgtm
/approve

PR #69, opened 2020-10-13
Fixed the build and push with skopeo

arjunrn:
`skopeo` expects a tag when given an image to push. This was missing. Also parameterized the build image tag.


2uasimojo:
Thanks @arjunrn. Looks right to me, but I guess we won't know it's working until we try it...

/lgtm
/approve

PR #71, opened 2020-10-13
Added variable for git hash

arjunrn:
Signed-off-by: Arjun Naik <anaik@redhat.com>


2uasimojo (approved):
...but we can clean that up later. Let's get this thing building.

/lgtm

PR #72, opened 2020-10-13
Consumers build/test with tagged images

2uasimojo:
The openshift/golang-osd-operator convention now syncs the image used by prow and appsre by editing build/Dockerfile and .ci-operator.yaml. The image tag used in both is the latest `image-v*` tag from the boilerplate repository.

PR #73, opened 2020-10-14
Fix bracing in push.sh

2uasimojo:
D'oh.

I copied the skopeo commands from [the `make` target](https://github.com/openshift/boilerplate/pull/72/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L30-L35) and I didn't notice that the variable interpolation was using parentheses rather than curly braces.

Fixed.


2uasimojo:
Well that failed exactly how it was supposed to.

Updated tag (which we generally wouldn't want to do, but sokay in this case.)

/test pr-check

2uasimojo:
> Updated tag (which we generally wouldn't want to do, but sokay in this case.)

This broke because test 04-update-from-master-and-revert's master now expects a tag to exist, and I had moved it.

So I'm actually forced to follow the proper process and create a new tag instead.

PR #74, opened 2020-10-15
Golangci-lint configuration can be overriden

rporres:
I expect different consumers of boilerplate to have different ideas of
what kind of linters / static code analyzers should be run via golangci.

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/standard.mk
+++ b/boilerplate/openshift/golang-osd-operator/standard.mk
@@ -47,6 +47,8 @@ GOBUILDFLAGS=-gcflags="all=-trimpath=${GOPATH}" -asmflags="all=-trimpath=${GOPAT
 # Relevant issue - https://github.com/golangci/golangci-lint/issues/734
 GOLANGCI_LINT_CACHE ?= /tmp/golangci-cache
 
+GOLANGCI_LINT_CONFIG ?= ${CONVENTION_DIR}/golangci.yml
+
 TESTTARGETS := $(shell ${GOENV} go list -e ./... | egrep -v "/(vendor)/")
 # ex, -v
 TESTOPTS :=
@@ -88,7 +90,7 @@ push: docker-push
 .PHONY: go-check
 go-check: ## Golang linting and other static analysis
 	${CONVENTION_DIR}/ensure.sh golangci-lint
-	GOLANGCI_LINT_CACHE=${GOLANGCI_LINT_CACHE} golangci-lint run -c ${CONVENTION_DIR}/golangci.yml ./...
+	GOLANGCI_LINT_CACHE=${GOLANGCI_LINT_CACHE} golangci-lint run -c ${GOLANGCI_LINT_CONFIG} ./...
 
 .PHONY: go-generate
 go-generate:


rporres:
Quick examples from a few openshift/app-sre projects:

* https://github.com/openshift/aws-account-operator/blob/master/.golangci.yml
* https://github.com/openshift/hive/blob/master/golangci.yml
* https://github.com/openshift/telemeter/blob/master/.golangci.yml
* https://github.com/app-sre/deployment-validation-operator/blob/master/.golangci.yml

arjunrn:
@2uasimojo please unhold when you are done reviewing

2uasimojo:
Hi there.

Part of the purpose of boilerplate, and the whole [standardization epic](https://issues.redhat.com/browse/SDE-669), is to minimize differences among consumers for anything that's not actual business logic. Linting falls squarely in that category.

Without going into a whole treatise on it, there's nontrivial value in having identical linting rules from one project to the next. Makes it easier for a developer to jump from one project to another and focus on content without having to learn the local dialect, maintain different IDE settings, etc.

Team Omega (owners of the standardization mission and this repo) have every intention to smooth the transition by helping consuming repositories with things like this. To that end, I went off to propose a PR making deployment-validation-operator lint-free per the existing config in this convention. And guess what!

```
[efried@efried deployment-validation-operator]$ make lint
boilerplate/openshift/golang-osd-operator/ensure.sh venv boilerplate/openshift/golang-osd-operator/py-requirements.txt
Requirement already satisfied: pyyaml>=5.3.1 in ./.venv/lib/python3.6/site-packages (from -r boilerplate/openshift/golang-osd-operator/py-requirements.txt (line 1))
You are using pip version 9.0.3, however version 20.2.3 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
.venv/bin/python3 boilerplate/openshift/golang-osd-operator/validate-yaml.py deploy/bundle/metadata/annotations.yaml deploy/bundle/template/deploymentvalidationoperator.clusterserviceversion.yaml deploy/openshift/cluster-role-binding.yaml deploy/openshift/cluster-role.yaml deploy/openshift/deployment-validation-operator-olm.yaml deploy/openshift/network-policies.yaml deploy/openshift/operator.yaml deploy/openshift/role-binding.yaml deploy/openshift/role.yaml deploy/openshift/service-account.yaml
Validating YAML deploy/bundle/metadata/annotations.yaml
Validating YAML deploy/bundle/template/deploymentvalidationoperator.clusterserviceversion.yaml
Validating YAML deploy/openshift/cluster-role-binding.yaml
Validating YAML deploy/openshift/cluster-role.yaml
Validating YAML deploy/openshift/deployment-validation-operator-olm.yaml
Validating YAML deploy/openshift/network-policies.yaml
Validating YAML deploy/openshift/operator.yaml
Validating YAML deploy/openshift/role-binding.yaml
Validating YAML deploy/openshift/role.yaml
Validating YAML deploy/openshift/service-account.yaml
boilerplate/openshift/golang-osd-operator/ensure.sh golangci-lint
/home/efried/.gvm/pkgsets/go1.13.6/global/bin/golangci-lint
GOLANGCI_LINT_CACHE=/tmp/golangci-cache golangci-lint run -c boilerplate/openshift/golang-osd-operator/golangci.yml ./...
[efried@efried deployment-validation-operator]$ echo $?
0
```

rporres:
The thing is golangci is much more than linting, it is a lot about static code analysis. Go is not very flexible in terms of code formatting since `go fmt` is enforced so I don't see the IDE problem here as much as I can see it in other languages. 

I don't envision a very easy way to agree between all the consumers about what are the checks that should be included in golanci as every project's needs are different. You can see the wild difference between different openshift projects and what boilerplate offers.

I don't understand the last part of your message, @2uasimojo . The fact that the code passes the boilerplate golangci checks config is expected as it contains less checks than deployment-validation-operator

2uasimojo:
> The thing is golangci is much more than linting, it is a lot about static code analysis. Go is not very flexible in terms of code formatting since `go fmt` is enforced so I don't see the IDE problem here as much as I can see it in other languages.

Right, I was overloading the term to include the static analysis performed by golangci-lint. IDEs (famously) do similar static analysis, including e.g. warning you about unused variables, missed error checks, etc. Most IDEs allow you to configure such analysis so you can identify and fix issues from within the IDE, rather than needing to run lint checks separately and switching back and forth. Unless properly configured, discrepancies between IDE configs and those run in CI can range from innocuous yellow squiggly underlines to actual blockers.

> I don't envision a very easy way to agree between all the consumers about what are the checks that should be included in golanci as every project's needs are different. You can see the wild difference between different openshift projects and what boilerplate offers.

Again, reducing those wild differences and promoting consistency across consumers is precisely our charter. Agree that consensus may be hard, but it's a challenge we're willing to take on :)

> I don't understand the last part of your message, @2uasimojo . The fact that the code passes the boilerplate golangci checks config is expected as it contains less checks than deployment-validation-operator

I may be dramatically misunderstanding how the configuration works, but [dvo's config](https://github.com/app-sre/deployment-validation-operator/blob/f0960a0945c4e94ec9d378c25402a02ee1c4e6d4/.golangci.yml) starts by [disabling all checks](https://github.com/app-sre/deployment-validation-operator/blob/f0960a0945c4e94ec9d378c25402a02ee1c4e6d4/.golangci.yml#L5) and then [enumerating an explicit list](https://github.com/app-sre/deployment-validation-operator/blob/f0960a0945c4e94ec9d378c25402a02ee1c4e6d4/.golangci.yml#L6-L28) of checks to include. Whereas [boilerplate's config](https://github.com/openshift/boilerplate/blob/7a12f7632155dc7b318eb91c463703640e47747d/boilerplate/openshift/golang-osd-operator/golangci.yml) has almost no customization, meaning that by default all checks are enabled. Am I missing something?

rporres:
>  IDEs (famously) do similar static analysis, including e.g. warning you about unused variables, missed error checks, etc. Most IDEs allow you to configure such analysis so you can identify and fix issues from within the IDE, rather than needing to run lint checks separately and switching back and forth. Unless properly configured, discrepancies between IDE configs and those run in CI can range from innocuous yellow squiggly underlines to actual blockers.

I don't understand very well what you're proposing here. The fact that we have a CI lint is precisely to avoid those IDE discrepancies.

> I may be dramatically misunderstanding how the configuration works, but dvo's config starts by disabling all checks and then enumerating an explicit list of checks to include. Whereas boilerplate's config has almost no customization, meaning that by default all checks are enabled. Am I missing something?

If I understand correctly what https://golangci-lint.run/usage/linters/ says, not all checks are enabled by default. Some of the checks included in dvo's config are not included in the default ones so more checks are being by dvo config as it is written. I see an advantage is disabling all checks and enabling them explicitly as what golangci-lint considers default can change with different versions.

> Again, reducing those wild differences and promoting consistency across consumers is precisely our charter. Agree that consensus may be hard, but it's a challenge we're willing to take on :)

We can have that conversation and agree in a list of checks for all consumers of boilerplate. That's hard but feasible. But since boilerplate is much more than golanci-lint, I see a value in letting people override this part while we get at that consensus and benefit from what boilerplate offers.

2uasimojo:
> I don't understand very well what you're proposing here. The fact that we have a CI lint is precisely to avoid those IDE discrepancies.

Sorry, I brought up IDEs only as an example of where consistency of linting rules can be advantageous. I can configure my IDE's rules to match what's enforced by my CI. But if I then switch to another project that uses _different_ CI configuration, I have to _reconfigure_ my IDE accordingly (or ignore the discrepancies, or whatever).

> If I understand correctly what https://golangci-lint.run/usage/linters/ says, not all checks are enabled by default. Some of the checks included in dvo's config are not included in the default ones so more checks are being run using dvo's config as it is written. I see an advantage is disabling all checks and enabling them explicitly as what golangci-lint considers default can change with different versions.

Okay, that's fair. As an exercise, I'll go try dvo's config on a couple of repositories I know to be using boilerplate's config (or something very similar) and see how far off they are. But this would be an argument for changing _boilerplate's_ published config, not for allowing arbitrary overrides per consumer. That said, we're also [standardizing the version of golangci-lint](https://github.com/openshift/boilerplate/blob/7a12f7632155dc7b318eb91c463703640e47747d/boilerplate/openshift/golang-osd-operator/ensure.sh#L7) being used, so we expect the defaults to drift (for everyone) when we change that version (explicitly and deliberately).

> We can have that conversation and agree in a list of checks for all consumers of boilerplate. That's hard but feasible. But since boilerplate is much more than golanci-lint, I see a value in letting people override this part while we get at that consensus and benefit from what boilerplate offers.

And the approach we're proposing is starting off with a config that's as generic as possible, applying it identically to all consumers, and then iterating on it from the source (boilerplate) as necessary. If we start off allowing bespoke configs per consumer, there is no motivation for a consumer to move to conformance, and we will never reach consistency.

2uasimojo:
> As an exercise, I'll go try dvo's config on a couple of repositories I know to be using boilerplate's config (or something very similar) and see how far off they are.

Did this on four repositories. You're right: there are plenty of new hits.

> But this would be an argument for changing _boilerplate's_ published config, not for allowing arbitrary overrides per consumer. 

Still prefer this, FWIW.

BUT...

Since the concern in this case is that you want to hold dvo to a higher standard, how would you feel about allowing bp consumers to specify a config that we will run _in addition_ to the configured default? That way consumers are at least held to the minimum standard, and are only allowed to to increase, not decrease, their strictness?

rporres:
> And the approach we're proposing is starting off with a config that's as generic as possible, applying it identically to all consumers, and then iterating on it from the source (boilerplate) as necessary. If we start off allowing bespoke configs per consumer, there is no motivation for a consumer to move to conformance, and we will never reach consistency.

I was thinking here in consumer's good faith, but I understand the motivation.

To have a standard set of checks may prove to be difficult when dealing with well stablished projects with large codebases...

rporres:
> Since the concern in this case is that you want to hold dvo to a higher standard, how would you feel about allowing bp consumers to specify a config that we will run in addition to the configured default? That way consumers are at least held to the minimum standard, and are only allowed to to increase, not decrease, their strictness?

I guess it's a fair compromise. I would write the boilerplate golangci configuration to be explicit in the checks instead of relying in the default config so that it is clearer to know which additional checks are being run.

2uasimojo:
> I guess it's a fair compromise. I would write the boilerplate golangci configuration to be explicit in the checks instead of relying in the default config so that it is clearer to know which additional checks are being run.

I can get behind that.

Are you volunteering to propose either or both of these pieces?
- Converting boilerplate's config to use the defaults explicitly instead of implicitly.
- Adding a second golangci-lint run to the `go-check` rule, triggered by and using a consumer config.

rporres:
Those shouldn't be difficult to implement. I will do them in the next days. In the meantime, I will close this PR as is.

2uasimojo:
> Those shouldn't be difficult to implement. I will do them in the next days.

Great, thanks @rporres! Of course reach out if you want to discuss/brainstorm.

PR #80, opened 2020-10-21
Make go-build only build

2uasimojo:
Since we're running them in prow, and have made sure the prow environment matches the appsre environment, there's no need for validation steps like linting and `go test` to be run as part of the container build.

This commit makes `make go-build` only do the go build. Since `make go-build` is the crux of `make docker-build`, this makes the container build more efficient.

(Note that, prior to #103, `make docker-build` was the only way to run these validation pieces "locally" in the proper container. Now you can do it with `boilerplate/_lib/container-make {target}`.)

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/standard.mk
+++ b/boilerplate/openshift/golang-osd-operator/standard.mk
@@ -64,7 +64,7 @@ CONVENTION_DIR := boilerplate/openshift/golang-osd-operator
 # https://www.gnu.org/software/make/manual/make.html#index-_002eDEFAULT_005fGOAL-_0028define-default-goal_0029
 .DEFAULT_GOAL :=
 .PHONY: default
-default: go-build
+default: go-check go-test go-build
 
 .PHONY: clean
 clean:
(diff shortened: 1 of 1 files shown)


boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -62,7 +62,7 @@ CONVENTION_DIR := boilerplate/openshift/golang-osd-operator
 # https://www.gnu.org/software/make/manual/make.html#index-_002eDEFAULT_005fGOAL-_0028define-default-goal_0029
 .DEFAULT_GOAL :=
 .PHONY: default
-default: go-build
+default: go-check go-test go-build
```
arjunrn, line 65: I'm no `make` expert but the general workflow I have seen is that simply running `make` just builds the binary/artifact. OTH running `make test` runs checks and other verification. So do we want to keep that workflow here as well. In other words should default just be `go-build`.

2uasimojo: I did actually ask about this a couple weeks ago, because I thought the same thing from my earliest experiences with `make`. I was told this was typical. Thinking of it empirically, IMO it makes sense for the default to be the case I would most commonly/frequently want to invoke as a developer, and this seems reasonable.

Anyway, happy to discuss further. This PR isn't going anywhere for a little while -- I'd like to hold it until we can reinstate a way to run check & test in the container.

2uasimojo:
/hold cancel

Now that we have #103, we can move forward with this.

PR #82, opened 2020-10-22
Image updates

2uasimojo:
This PR makes two changes to the image build config:

- Add a comment to explain why we're building git from source (nonfunctional)
- Install SSH clients

### DIFF
--- a/config/build.sh
+++ b/config/build.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -x
 set -euo pipefail
 
 tmpd=$(mktemp -d)
--- a/config/build.sh
+++ b/config/build.sh
@@ -31,6 +31,11 @@ mv opm /usr/local/bin
 
 python3 -m pip install PyYAML==5.3.1
 
+# Per https://git-scm.com/download/linux, we have two choices for CentOS
+# (which is what we're running on):
+# - Build from source
+# - Use a third party repository
+# For security reasons, we're preferring the former.
 GIT_VERSION="2.28.0"
 GIT_SHASUM="02016d16dbce553699db5c9c04f6d13a3f50727c652061b7eb97a828d045e534"
 GIT_DEPENDENCIES="epel-release perl-CPAN gettext-devel perl-devel openssl-devel zlib-devel curl-devel expat-devel getopt asciidoc xmlto docbook2X"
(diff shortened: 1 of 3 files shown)


config/build.sh
```diff
@@ -23,14 +23,21 @@ mv golangci-lint-${GOCILINT_VERSION}-linux-amd64/golangci-lint /usr/local/bin
...
 python3 -m pip install PyYAML==5.3.1
 
+# Per https://git-scm.com/download/linux, we have two choices for CentOS
+# (which is what we're running on):
+# - Build from source
+# - Use a third party repository
+# For security reasons, we're preferring the former.
 GIT_VERSION="2.28.0"
 GIT_SHASUM="02016d16dbce553699db5c9c04f6d13a3f50727c652061b7eb97a828d045e534"
 GIT_DEPENDENCIES="epel-release perl-CPAN gettext-devel perl-devel openssl-devel zlib-devel curl-devel expat-devel getopt asciidoc xmlto docbook2X"
```
dustman9000, line 43: I think we need to break apart this variable into `BUILD_DEPENDENCIES` and `RUNTIME_DEPENDENCES`. When we remove `curl-devel` and `openssl-devel` we are removing core packages depended by ssh and other core utils.

config/build.sh
```diff
@@ -47,6 +54,12 @@ yum groupremove -y "Development Tools" && \
 yum -y remove ${GIT_DEPENDENCIES}
 yum clean all
 yum -y autoremove
+
+# autoremove removes ssh (which it presumably wouldn't if we were able
+# to install git from a repository, because git has a dep on ssh.)
+# Do we care to restrict this to a particular version?
+yum -y install openssh-clients
```
dustman9000, line 61: Let's create a new variable called something like RUNTIME_DEPENDENCES and add `curl`, `openssh-clients` and other packages removed by the call that removes all the `*-devel` packages.

config/build.sh
2uasimojo: Per offline discussion, we determined these were sane build deps not affecting the real `curl` or `ssh`. So leaving this alone.

config/build.sh
2uasimojo: ditto.

2uasimojo:
Merge commit wants the tag. Trivial rebase. Will reapply lgtm if CI is happy.

PR #90, opened 2020-10-28
Skip operator-sdk generate if no APIs

2uasimojo:
If your operator has no APIs (which is a legitimate thing), `operator-sdk generate` will fail with a message like:

`error generating CRDs from APIs in pkg/apis: error generating CRD manifests: error reading CRD cache dir deploy/crds: open deploy/crds: file does not exist`

Add a check to operator-sdk-generate.sh to skip if no APIs are found.


arjunrn:
:+1:

2uasimojo:
Arjun's lgtm didn't stick. Adding it manually.

PR #91, opened 2020-10-30
Make default golangci-lint config rules explicit

rporres:
We will know better which checks are in use and they won't change if
defaults are changed between golangci-lint versions

Addresses [OSD-5557](https://issues.redhat.com/browse/OSD-5557)


rporres:
Defaults taken from https://golangci-lint.run/usage/linters/#enabled-by-default-linters

PR #92, opened 2020-10-30
Optional second golangci-lint config

rporres:
This will be controlled by `GOLANGCI_OPTIONAL_CONFIG` and it will allow
boilerplate consumers to run additional checks not covered by the
default config

Addresses [OSD-5558](https://issues.redhat.com/browse/OSD-5558)

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/standard.mk
+++ b/boilerplate/openshift/golang-osd-operator/standard.mk
@@ -47,6 +47,8 @@ GOBUILDFLAGS=-gcflags="all=-trimpath=${GOPATH}" -asmflags="all=-trimpath=${GOPAT
 # Relevant issue - https://github.com/golangci/golangci-lint/issues/734
 GOLANGCI_LINT_CACHE ?= /tmp/golangci-cache
 
+GOLANGCI_OPTIONAL_CONFIG ?=
+
 TESTTARGETS := $(shell ${GOENV} go list -e ./... | egrep -v "/(vendor)/")
 # ex, -v
 TESTOPTS :=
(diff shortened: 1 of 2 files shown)


boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -89,6 +91,7 @@ push: docker-push
 go-check: ## Golang linting and other static analysis
 	${CONVENTION_DIR}/ensure.sh golangci-lint
 	GOLANGCI_LINT_CACHE=${GOLANGCI_LINT_CACHE} golangci-lint run -c ${CONVENTION_DIR}/golangci.yml ./...
+	test "${GOLANGCI_OPTIONAL_CONFIG}" = "" || test ! -e "${GOLANGCI_OPTIONAL_CONFIG}" || GOLANGCI_LINT_CACHE="${GOLANGCI_LINT_CACHE}" golangci-lint run -c "${GOLANGCI_OPTIONAL_CONFIG}" ./...
```
2uasimojo, line 94: This looks right.

Maybe it's just me, but I find the logic confusing as written. To me it would be more intuitive as a chain of positive assertions:
```suggestion
	test -n "${GOLANGCI_OPTIONAL_CONFIG}" && test -f "${GOLANGCI_OPTIONAL_CONFIG}" && GOLANGCI_LINT_CACHE="${GOLANGCI_LINT_CACHE}" golangci-lint run -c "${GOLANGCI_OPTIONAL_CONFIG}" ./...
```

2uasimojo (approved):
Thanks @rporres!

/lgtm

Optional suggestion inline. Fix if you like, or unhold to ignore.

/hold

boilerplate/openshift/golang-osd-operator/standard.mk
rporres: It would definitely be much more intuitive but unfortunately the target will return `1` from the target if `GOLANGCI_OPTIONAL_CONFIG` is not defined or doesn't exist :(

2uasimojo: Oh, yeah.

PR #98, opened 2020-11-18
Glue to CRD v1beta1

2uasimojo:
Support for generating v1 CustomResourceDefinitions was introduced in operator-sdk 0.17, and in 0.18 it is the default. OSD operators need to be cautious about cutting over to v1 as it is impacted by which cluster versions are supported, and by which operators are deployed to which clusters. So for now, we're forcing `operator-sdk generate crds` to use CRD v1beta1.


2uasimojo:
/hold cancel

Works.

PR #100, opened 2020-11-23
Add support for grpcurl and opm in ensure.sh

rporres:
This is in preparation of the catalog creation script using `opm` that will be added to boilerplate.


2uasimojo (approved):
Yah, nice.

/lgtm
/approve

PR #101, opened 2020-11-25
First commit for the bundle-generate

bdematte:


boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
```diff
@@ -0,0 +1,121 @@
...
+    operator_deployment = operator_components[0]
+    csv['spec']['install']['spec']['deployments'][0]['spec'] = operator_deployment['spec']
+
+# Update the deployment to use the defined image:
+csv['spec']['install']['spec']['deployments'][0]['spec']['template']['spec']['containers'][0]['image'] = operator_image
+
+# Update the versions to include git hash:
+csv['metadata']['name'] = "{}.v{}".format(operator_name, full_version)
+csv['spec']['version'] = full_version
+csv['spec']['replaces'] = "{}.v{}".format(operator_name, prev_version)
```
jharrington22, line 110: We need a way to not write this when this is the first deploy as there will be no "replaces".

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
```diff
@@ -0,0 +1,118 @@
+#!/bin/bash -x
```
jharrington22, line 1: This would replace app_sre_build_image_catalog.sh?

test/projects/file-generate/hack/generate-operator-bundle.py
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env python
```
jharrington22, line 1: How do we keep these files in sync?

jharrington22 (commented):
Submodule? test/projects/saas-file-generate-bundle

jharrington22:
@bdematte could you look at integrating validation of the built catalog image https://github.com/openshift/osd-metrics-exporter/pull/32

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
bdematte: The 3 shell scripts form `csv-generate` folder are replacing altogether the `app_sre_build_image_catalog.sh`

bdematte:
> Submodule? test/projects/saas-file-generate-bundle

No, this is just a folder used to seed a `saas-file-generate-bundle` temporary git repo for the testing, but the git init is part of the test case itself.

test/projects/file-generate/hack/generate-operator-bundle.py
bdematte: which files ?

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
```diff
@@ -0,0 +1,118 @@
...
+                fi
+            else
+                rm -rf "${BUNDLE_DIR:?BUNDLE_DIR var not set}/$version"
+            fi
+        done
+    fi
+    
+    OPERATOR_PREV_VERSION=$(ls "$BUNDLE_DIR" | sort -t . -k 3 -g | tail -n 1)
+    # CSV actual generate
+    ./hack/generate-operator-bundle.py -o ${operator_name} -d ${BUNDLE_DIR} -p ${OPERATOR_PREV_VERSION} -n ${operator_commit_number} -c ${operator_commit_hash} -i ${operator_image}
```
bdematte, line 116: I think the script is not taking options but ordered parameters in other operators - to be checked and fixed

boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
rporres: This is what I did in dvo: https://github.com/app-sre/deployment-validation-operator/blob/master/hack/generate-operator-bundle-contents.py#L72-L73

But we have to be sure that the CSV template does not contain a `replaces` field

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
bdematte: Fixed in latest commit

bdematte:
Recap : 
  - Added support for onboarding new component (ie no previous-version)
To be addressed in follow-up PR : 
  - Add check on the generated image similar to openshift/osd-metrics-exporter#32 
  - Add support for long-named options

test/projects/saas-file-generate-bundle/file-generate/file-generate.package.yaml
```diff
@@ -0,0 +1,4 @@
+packageName: file-generate
+channels:
+- name: staging
+  currentCSV: file-generate.v0.1.10-c664c72
```
jharrington22, line 4: The test case creates a new version `0.1.6-441cd20` do we validate that this is updated?

test/projects/file-generate/hack/generate-operator-bundle.py
jharrington22: Ah I see, this is purely a test project.

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -161,6 +161,75 @@ prow-config:
...
+.PHONY: production-common-csv-build
+production-common-csv-build: 
+	@${CONVENTION_DIR}/csv-generate/csv-generate.sh -o $(OPERATOR_NAME) -i $(OPERATOR_IMAGE) -c production -h $(CURRENT_COMMIT) -n $(COMMIT_NUMBER) -g common
+
+.PHONY: production-common-csv-build-and-diff
+production-common-csv-build-and-diff: 
+	@${CONVENTION_DIR}/csv-generate/csv-generate.sh -o $(OPERATOR_NAME) -i $(OPERATOR_IMAGE) -c production -h $(CURRENT_COMMIT) -n $(COMMIT_NUMBER) -g common -d
+
+.PHONY: production-catalog-build
+production-catalog-saas-build:
```
jharrington22, line 216: Is this name correct?

jharrington22 (changes requested):
@bdematte this looks good I had a couple of small comments and some questions below:

By default operators wont use the common generate script?

Were we going to validate the catalog image as part of `catalog-build.sh` ?

boilerplate/openshift/golang-osd-operator/csv-generate/catalog-build.sh
```diff
@@ -0,0 +1,68 @@
...
+    csv_missing_param_error = true
+fi
+
+if [ ! -z "$csv_missing_param_error" ] ; then
+    usage
+    exit 1
+fi
+
+# If no override, using the gitlab repo
+if [ -z "$GIT_PATH" ] ; then
```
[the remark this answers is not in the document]
bdematte: The only case of override is the testing, that's why I didn't want to make it as an explicit parameter

bdematte:
> @bdematte this looks good I had a couple of small comments and some questions below:
> 
> By default operators wont use the common generate script?
Nope, by default, operators will continue to use the generate script from `hack` folder to avoid unneeded change. However, operators having a generate script which doesn't support the same parameters than the majority will have the ability to use the common generate script to avoid unneeded adaptation. 
 
> Were we going to validate the catalog image as part of `catalog-build.sh` ?

Yes, I put the todo to track it. Idea is to check what is done in `metric-exporter` and for the 'new push' model and integrate it in here at the end of the build (so that people aiming at local deployment with olm in the future can take advantage of it)

boilerplate/openshift/golang-osd-operator/csv-generate/catalog-build.sh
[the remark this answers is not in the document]
jharrington22: Ok makes sense

boilerplate/openshift/golang-osd-operator/csv-generate/catalog-publish.sh
```diff
@@ -0,0 +1,101 @@
...
+    csv_missing_param_error=true
+fi
+
+if [ ! -z "$csv_missing_param_error" ] ; then
+    usage
+    exit 1
+fi
+
+# If no override, using the gitlab repo
+if [ -z "$GIT_PATH" ] ; then
```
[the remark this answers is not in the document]
jharrington22: See above, only overwritten in testing.

boilerplate/openshift/golang-osd-operator/csv-generate/catalog-build.sh
```diff
@@ -0,0 +1,67 @@
...
+fi
+
+if [ -z "$operator_commit_hash" ] ; then 
+    echo "Missing operator_commit_hash parameter"
+    csv_missing_param_error = true
+fi
+
+if [ ! -z "$csv_missing_param_error" ] ; then
+    usage
+fi
```
2uasimojo, line 41: Since you're using this pattern a lot for mandatory params, you could DRY it up with a common function like:

```bash
function check_mandatory_params() {
    local csv_missing_param_error
    local param_name
    local param_val
    for param_name in "$@"; do
        eval param_val=\$$param_name
        if [ -z "$param_val" ]; then
            echo "Missing $param_name parameter"
            csv_missing_param_error=true
        fi
    done
    if [ ! -z "$csv_missing_param_error" ]; then
        usage
    fi
}
```

Then just e.g.

```bash
check_mandatory_params operator_channel operator_name operator_commit_hash
```

boilerplate/openshift/golang-osd-operator/csv-generate/catalog-publish.sh
```diff
@@ -0,0 +1,99 @@
+#!/bin/bash -x
+
+usage() { echo "Usage: $0 -o operator_name -i operator_image -c saas-repository-channel [-t]" 1>&2; exit 1; }
```
2uasimojo, line 3: Fix to match actual CLI spec.

boilerplate/openshift/golang-osd-operator/csv-generate/catalog-publish.sh
```diff
@@ -0,0 +1,99 @@
...
+
+while getopts "o:c:n:h:p" option; do
+    case "${option}" in
+        c)
+            operator_channel=${OPTARG}
+            ;;
+        h)
```
2uasimojo, line 10: See note above about using `-h`.

boilerplate/openshift/golang-osd-operator/csv-generate/catalog-publish.sh
```diff
@@ -0,0 +1,99 @@
...
+fi
+
+if [ ! -z "${csv_missing_param_error}" ] ; then
+    usage
+fi
+
+# If no override, using the gitlab repo
+if [ -z "${GIT_PATH}" ] ; then 
+    GIT_PATH="https://app:@gitlab.cee.redhat.com/service/saas-${operator_name}-bundle.git"
+fi
```
2uasimojo, line 57: Unused?

boilerplate/openshift/golang-osd-operator/csv-generate/catalog-publish.sh
```diff
@@ -0,0 +1,99 @@
...
+EOF
+
+# add, commit & push
+pushd ${SAAS_OPERATOR_DIR}
+
+git add .
+
+MESSAGE="add version ${operator_commit_number}-${operator_commit_hash}
+replaces ${OPERATOR_PREV_VERSION}
+removed versions: ${REMOVED_VERSIONS}"
```
2uasimojo, line 80: Do you need an extra newline after the header?
```suggestion
MESSAGE="add version ${operator_commit_number}-${operator_commit_hash}

replaces ${OPERATOR_PREV_VERSION}
removed versions: ${REMOVED_VERSIONS}"
```

boilerplate/openshift/golang-osd-operator/csv-generate/catalog-publish.sh
```diff
@@ -0,0 +1,99 @@
...
+# If no override, using the gitlab repo
+if [ -z "${GIT_PATH}" ] ; then 
+    GIT_PATH="https://app:@gitlab.cee.redhat.com/service/saas-${operator_name}-bundle.git"
+fi
+
+# Calculate previous version
+SAAS_OPERATOR_DIR="saas-${operator_name}-bundle"
+BUNDLE_DIR="${SAAS_OPERATOR_DIR}/${operator_name}/"
+OPERATOR_NEW_VERSION=$(ls "${BUNDLE_DIR}" | sort -t . -k 3 -g | tail -n 1)
+OPERATOR_PREV_VERSION=$(ls "${BUNDLE_DIR}" | sort -t . -k 3 -g | tail -n 2 | head -n 1)
```
2uasimojo, line 63: Ditto above about assuming the clone exists, and about error checking.

boilerplate/openshift/golang-osd-operator/csv-generate/catalog-publish.sh
```diff
@@ -0,0 +1,99 @@
...
+replaces ${OPERATOR_PREV_VERSION}
+removed versions: ${REMOVED_VERSIONS}"
+
+git commit -m "${MESSAGE}"
+git push origin "${operator_channel}"
+
+popd
+
+if [ "$push_catalog" = true ] ; then
+    # build the registry image
```
2uasimojo, line 88: bad comment

boilerplate/openshift/golang-osd-operator/csv-generate/catalog-publish.sh
```diff
@@ -0,0 +1,99 @@
...
+fi
+
+# If no override, using the gitlab repo
+if [ -z "${GIT_PATH}" ] ; then 
+    GIT_PATH="https://app:@gitlab.cee.redhat.com/service/saas-${operator_name}-bundle.git"
+fi
+
+# Calculate previous version
+SAAS_OPERATOR_DIR="saas-${operator_name}-bundle"
+BUNDLE_DIR="${SAAS_OPERATOR_DIR}/${operator_name}/"
```
2uasimojo, line 61: nit:
```suggestion
BUNDLE_DIR="${SAAS_OPERATOR_DIR}/${operator_name}"
```
otherwise you're doubling slashes below, which is harmless, but kind of icky.

boilerplate/openshift/golang-osd-operator/csv-generate/catalog-publish.sh
```diff
@@ -0,0 +1,99 @@
...
+pushd ${SAAS_OPERATOR_DIR}
+
+git add .
+
+MESSAGE="add version ${operator_commit_number}-${operator_commit_hash}
+replaces ${OPERATOR_PREV_VERSION}
+removed versions: ${REMOVED_VERSIONS}"
+
+git commit -m "${MESSAGE}"
+git push origin "${operator_channel}"
```
2uasimojo, line 83: Do these merge automatically?

boilerplate/openshift/golang-osd-operator/csv-generate/catalog-publish.sh
```diff
@@ -0,0 +1,99 @@
...
+
+popd
+
+if [ "$push_catalog" = true ] ; then
```
2uasimojo, line 87: It seems weird to have these two pieces (update the saas bundle repo, push the image built externally by catalog-build.sh) in the same script, but the latter is optional. I get that you're trying to make everything modular, presumably for testing purposes; so it seems like separating these two would also make sense. That could also simplify some of the make target spread.

2uasimojo (commented):
Partial review.

In general, I like where this is headed.

test/case/convention/openshift/golang-osd-operator/06-csv-generate
```diff
@@ -0,0 +1,105 @@
...
+popd
+
+DEPLOYED_HASH="0.1.0-441cd20"
+
+cd $repo
+
+make boilerplate-update
+
+echo -e "\e[33m First build and compare on Staging"
+make staging-hack-csv-build
```
2uasimojo, line 47: Won't this try to clone the saas repo for `file-generate`?

test/case/convention/openshift/golang-osd-operator/06-csv-generate
```diff
@@ -0,0 +1,105 @@
...
+make staging-common-csv-build-and-diff
+
+# Something to be checked/tested here
+
+echo -e "\e[33m First push on Staging"
+make staging-saas-bundle-push
```
2uasimojo, line 53: ...and won't this try to update it, and push an image?

How is `pr-check` succeeding in CI?

test/lib.sh
```diff
@@ -347,3 +347,15 @@ expect_failure () {
...
     return 0
 }
+
+
+## update_project PROJECT
+#
+# Generate dummy commit to a git project and commit it
+# :param PROJECT: A git ptoject 
+local_dummy_commit () {
+    new_file=`cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 10 | head -n 1`
```
2uasimojo, line 357: Yeek, why not just 
```suggestion
    new_file=`mktemp -p.`
```

2uasimojo (commented):
Okay, more or less done reviewing for now.

Mainly, I don't understand how the test case is working. If it's not (if it's passing by chance, e.g. because failures aren't being detected), IMHO it should be fixed up before we try to land this; otherwise we'll have a heck of a time migrating consumers.

Though even then, I don't see being able to do convincing "local" tests without setting up a dummy saas repo and quay project. We ought to do that at least for manual testing, and then dry run this with a real consumer or two.

boilerplate/openshift/golang-osd-operator/csv-generate/catalog-build.sh
```diff
@@ -0,0 +1,67 @@
...
+
+while getopts "o:h:c:" option; do
+    case "${option}" in
+        o)
+            operator_name=${OPTARG}
+            ;;
+        c)
+            operator_channel=${OPTARG}
+            ;;
+        h)
```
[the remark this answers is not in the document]
bdematte: As this is already evaluated within the make file (for opm-build-push), I believe it is better to use everywhere that single value rather than having it evaluated (potentially to different values) within multiple scripts (may it be from a common lib or not).

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -161,6 +161,75 @@ prow-config:
 .PHONY: codecov-secret-mapping
 codecov-secret-mapping:
 	${CONVENTION_DIR}/codecov-secret-mapping ${RELEASE_CLONE}
+
```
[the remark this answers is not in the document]
bdematte: I considered making a dedicated file and forgot to raise the discussion in the team (to agree on the generic strategy). 
I believe none of them are useless and using env variable will only make it more complex to debug/use to me. However, some of them are already planned to disappear once migrated to the common CSV generator and/or new versioning model.

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -161,6 +161,75 @@ prow-config:
...
 codecov-secret-mapping:
 	${CONVENTION_DIR}/codecov-secret-mapping ${RELEASE_CLONE}
+	
+.PHONY: staging-csv-build
+staging-csv-build: 
+	@${CONVENTION_DIR}/csv-generate/csv-generate.sh -o $(OPERATOR_NAME) -i $(OPERATOR_IMAGE) -c staging -h $(CURRENT_COMMIT) -n $(COMMIT_NUMBER) -g hack
+	
+.PHONY: staging-hack-csv-build
+staging-hack-csv-build: 
+	@${CONVENTION_DIR}/csv-generate/csv-generate.sh -o $(OPERATOR_NAME) -i $(OPERATOR_IMAGE) -c staging -h $(CURRENT_COMMIT) -n $(COMMIT_NUMBER) -g hack
```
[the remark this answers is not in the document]
bdematte: It is intentional, the goal being to allow user to override the default generator (which will be switched eventually to the common generator) in order not to be impacted by boilerplate default switch. 
But yes, I updated just to call the former target.

boilerplate/openshift/golang-osd-operator/csv-generate/catalog-publish.sh
bdematte: Those are git repos with allowed direct push, so no MR/Merge needed

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
```diff
@@ -0,0 +1,148 @@
...
+    case "${option}" in
+        c)
+            operator_channel=${OPTARG}
+            ;;
+        d)
+            diff_generate=true
+            ;;
+        g)
+            generate_script=${OPTARG}
+            ;;
```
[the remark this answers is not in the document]
bdematte: Script is not expected to be used manually (it can but not the standard usecase which is to use make target) and within code, I would prefer avoiding the default - still, I will add check on the values

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
```diff
@@ -0,0 +1,148 @@
...
+
+if [[ "$generate_script" = "common" ]] ; then
+    python3 ./boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py -o ${operator_name} -d ${OUTPUT_DIR} -p ${OPERATOR_PREV_VERSION} -n ${operator_commit_number} -c ${operator_commit_hash} -i ${operator_image}
+elif [[ "$generate_script" = "hack" ]] ; then
+    if [ -z "$OPERATOR_PREV_VERSION" ] ; then 
+        OPERATOR_PREV_VERSION="no-version"
+        DELETE_REPLACE=true
+    fi
+    
+    python3 ./hack/generate-operator-bundle.py ${OUTPUT_DIR} ${OPERATOR_PREV_VERSION} ${operator_commit_number} ${operator_commit_hash} ${operator_image}
```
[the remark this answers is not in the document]
bdematte: I believe they are not py2-specific (or it depends on operator maybe) as the common script is largely derivated from existing operators. Anyway, it makes sense to delegate it to the script itself indeed - I will update accodingly and yes, this is part of the point being covered in the fup PR with docs

test/case/convention/openshift/golang-osd-operator/06-csv-generate
bdematte: The GIT_PATH is overriden to local path which contains a brand new temp saas repo for the file-generate.
This allows to be independant from gitlab and have a stable/controlled saas-repo for the testing without any dependency.

test/case/convention/openshift/golang-osd-operator/06-csv-generate
bdematte: This is the reason why I added the parameter to have the option not to push the image to the registry (and we are using local git repo so it is updating a repo which is dedicated to this run of the test case)

boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
```diff
@@ -0,0 +1,121 @@
...
+# (presumably since some past base tag), and the git hash arg for a final
+# version like: 0.1.189-3f73a592
+VERSION_BASE = "0.1"
+
+parser = argparse.ArgumentParser()
+parser.add_argument("-o", "--operator-name", type=str, help="Name of the operator", required=True)
+parser.add_argument("-d", "--output-dir", type=str, help="Directory for the CSV generation", required=True)
+parser.add_argument("-p", "--previous-version", type=str, help="Directory for the CSV generation", required=True)
+parser.add_argument("-n", "--commit-number", type=str, help="Number of commits in the project (used for version generation)", required=True)
+parser.add_argument("-c", "--commit-hash", type=str, help="Current commit hashDirectory for the CSV generation (used for version generation)", required=True)
```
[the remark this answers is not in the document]
bdematte: Those values are calculated within Makefile for the work done for dvo . If we want to move it, I would make it a dedicated card to ensure we are using the same values across the board.

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
```diff
@@ -0,0 +1,148 @@
...
+    
+                if [[ "$DEPLOYED_HASH" == "${short_hash}"* ]]; then
+                    delete=true
+                fi
+            else
+                rm -rf "${BUNDLE_DIR:?BUNDLE_DIR var not set}/$version"
+            fi
+        done
+    fi
+    # TODO : Clean handling of major version (should be variable from consumer repo and default to 0.1 is undefined)
```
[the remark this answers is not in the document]
bdematte: The objective is to remove it from scripts when we'll do so (hence it is a TO DO as more thinking needed if we want to do it without forcing usage of the common script).

bdematte:
Open points : 

- For the hash calculation and commit count, do we want to use script parameters or shared lib to be called my each script ? I am personally in favour of being aligned with dvo work (values calculated in Makefile) and to eventually move all (legacy and new model scripts) to shared lib in separated MR if this is the direction we want to go.
- split 'catalog-publish.sh' into 2 scripts (1 for the git push and 1 for the image push) rather than having an option for it ? Outside of testing, there is no reason to use that option (used to be able to test without pushing to the central registry). If prefered, we can make it an env variable to make the interface of the script cleaner. 

Points for follow-up MR : 

- Documentation (how to use those target, how to migrate, how to run tests within CI container for debugging, ...)
- Add checks in the script(s) (check validity of the generated image, check SAAS directory is using proper remote, ..._
- Improve tests (check CSV version is updated in generated CSV, test if possible push to registry and docker build)
- Support for long options

boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
```diff
@@ -0,0 +1,111 @@
...
```
2uasimojo, line 27:
```suggestion
parser.add_argument("-c", "--commit-hash", type=str, help="Current commit hash (used for version generation)", required=True)
```

boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
```diff
@@ -0,0 +1,111 @@
...
+parser.add_argument("-i", "--operator-image", type=str, help="Base index image to be used", required=True)
+args = parser.parse_args()
+
+operator_name   = args.operator_name
+outdir          = args.output_dir
+prev_version    = args.previous_version
+git_num_commits = args.commit_number
+git_hash        = args.commit_hash
+operator_image  = args.operator_image
```
2uasimojo, line 36: nit: not sure why we need extra vars for these.

boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
```diff
@@ -0,0 +1,111 @@
...
+import sys
+import yaml
+import shutil
+import argparse
+import string
+
+# This script will append the current number of commits given as an arg
+# (presumably since some past base tag), and the git hash arg for a final
+# version like: 0.1.189-3f73a592
+VERSION_BASE = "0.1"
```
2uasimojo, line 20: If we're running this script via the Makefile, there are [env vars](https://github.com/openshift/boilerplate/blob/69bf49e355dd9fe677a6c2d7a295949f6a2998e7/boilerplate/openshift/golang-osd-operator/project.mk#L9-L10) for this major/minor, which may be [overridden by some consumers](https://github.com/openshift/managed-velero-operator/blob/7ecdc226fd2d716c7d91de6cd0a6a712cefbc6e9/Makefile#L3). We should really use those instead of hardcoding.

boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
```diff
@@ -0,0 +1,111 @@
...
+
+full_version = "%s.%s-%s" % (VERSION_BASE, git_num_commits, git_hash)
+print("Generating CSV for version: %s" % full_version)
+
+if not os.path.exists(outdir):
+    os.mkdir(outdir)
+
+version_dir = os.path.join(outdir, full_version)
+if not os.path.exists(version_dir):
+    os.mkdir(version_dir)
```
2uasimojo, line 46: FYI, there's an [`os.makedirs()`](https://docs.python.org/3/library/os.html#os.makedirs) that creates directories recursively. So you could:

```suggestion
version_dir = os.path.join(outdir, full_version)
if not os.path.exists(version_dir):
    os.makedirs(version_dir)
```

boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
```diff
@@ -0,0 +1,111 @@
...
+
+with open('config/templates/csv-template.yaml'.format(operator_name), 'r') as stream:
```
2uasimojo, line 48: Did you mean to use a substitution placeholder (`{}`) in the format string somewhere? As written, the `.format(operator_name)` part of this is a no-op.

[Later] Based on what I see elsewhere, I think you meant to leave out the operator name bit.
```suggestion
with open('config/templates/csv-template.yaml', 'r') as stream:
```

boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
2uasimojo: This is where we would want it to be legal to run with `-p ""` or some other signal of "no previous version".

In that case, we could either make sure `replaces` is absent from the template and add it here if necessary, as you suggest; or we could just delete it here.

```suggestion
if prev_version:
    csv['spec']['replaces'] = "{}.v{}".format(operator_name, prev_version)
else:
    del csv['spec']['replaces']
```

Or more carefully (works even if `replaces` isn't in the template):

```suggestion
if prev_version:
    csv['spec']['replaces'] = "{}.v{}".format(operator_name, prev_version)
else:
    csv.pop(['spec']['replaces'], None)
```

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
```diff
@@ -0,0 +1,127 @@
+#!/bin/bash -e
+
+source `dirname $0`/common.sh
+
+usage() { echo "Usage: $0 -o operator_name -c saas-repository-channel -H operator_commit_hash -n operator_commit_number -i operator_image -g [hack|common][-d]" 1>&2; exit 1; }
```
2uasimojo, line 5: Use curly braces for mandatory enum options
```suggestion
usage() { echo "Usage: $0 -o operator_name -c saas-repository-channel -H operator_commit_hash -n operator_commit_number -i operator_image -g {hack|common} [-d]" 1>&2; exit 1; }
```

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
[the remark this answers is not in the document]
2uasimojo: As noted above, we ought to have env vars available for this.

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
```diff
@@ -0,0 +1,127 @@
...
+                if [[ "$DEPLOYED_HASH" == "${short_hash}"* ]]; then
+                    delete=true
+                fi
+            else
+                rm -rf "${BUNDLE_DIR:?BUNDLE_DIR var not set}/$version"
+            fi
+        done
+    fi
+    # TODO : Clean handling of major version (should be variable from consumer repo and default to 0.1 is undefined)
+    OPERATOR_PREV_VERSION=$(ls "$BUNDLE_DIR" | sort -t . -k 3 -g | tail -n 1)
```
2uasimojo, line 94: Again here we should account for the case where we're building the first version.

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
```diff
@@ -0,0 +1,127 @@
...
+    OPERATOR_NEW_VERSION="0.1.${operator_commit_number}-${operator_commit_hash}"
+    OUTPUT_DIR=${BUNDLE_DIR}
+fi
+
+if [[ "$generate_script" = "common" ]] ; then
+    ./boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py -o ${operator_name} -d ${OUTPUT_DIR} -p ${OPERATOR_PREV_VERSION} -n ${operator_commit_number} -c ${operator_commit_hash} -i ${operator_image}
```
2uasimojo, line 100: It would be cool to compute this path rather than hardcoding it.
```suggestion
    HERE=${0%/*}
    $HERE/common-generate-operator-bundle.py -o ${operator_name} -d ${OUTPUT_DIR} -p ${OPERATOR_PREV_VERSION} -n ${operator_commit_number} -c ${operator_commit_hash} -i ${operator_image}
```

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
```diff
@@ -0,0 +1,127 @@
...
+elif [[ "$generate_script" = "hack" ]] ; then
+    if [ -z "$OPERATOR_PREV_VERSION" ] ; then 
+        OPERATOR_PREV_VERSION="no-version"
+        DELETE_REPLACE=true
+    fi
+    
+    ./hack/generate-operator-bundle.py ${OUTPUT_DIR} ${OPERATOR_PREV_VERSION} ${operator_commit_number} ${operator_commit_hash} ${operator_image}
+    
+    if [ ! -z "${DELETE_REPLACE}" ] ; then
+        yq d -i output-comparison/${OPERATOR_NEW_VERSION}/*.clusterserviceversion.yaml 'spec.replaces'
```
2uasimojo, line 110: FYI we've had trouble with different versions of `yq` being present wherever this runs. v3 and v4 have wildly different and incompatible syntax.

But if we detect this condition in generate-operator-bundle.py and delete `spec.replaces` there as suggested above, this chunk goes away and we don't have to worry about it :)

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
```diff
@@ -0,0 +1,127 @@
...
+    fi
+fi
+
+if [ "$diff_generate" = true ] ; then
+    # TODO : Current hack script does not allow to generate the CSV for the comparison (it will generate a different version that the common one because there is 1 extra version in the history)
+    if [[ "$generate_script" = "hack" ]] ; then
+        echo "Generating with the common script and after, generating with the hack script is not supported yet. For comparison, please first generate with hack script, and then build/compare with the common script"
+        exit 1
```
2uasimojo, line 118: If we get here, we've already generated the bundle, which leaves a mess. This seems like it would be an easy condition to check early, before we've written anything to disk.

```bash
if [[ "$diff_generate" == "true" ]] && [[ "$generate_script" == "hack" ]]; then
    err "Generating .... is not supported ..."
fi
```

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
```diff
@@ -0,0 +1,127 @@
...
+    # Preparing yamls for the diff by removing the creation timestamp
+    elif [ -f ${BUNDLE_DIR}/${OPERATOR_NEW_VERSION}/*.clusterserviceversion.yaml ] ; then
+        yq d ${BUNDLE_DIR}/${OPERATOR_NEW_VERSION}/*.clusterserviceversion.yaml 'metadata.annotations.createdAt' > output-comparison/hack_generate.yaml
+        yq d output-comparison/${OPERATOR_NEW_VERSION}/*.clusterserviceversion.yaml 'metadata.annotations.createdAt' > output-comparison/common_generate.yaml
```
2uasimojo, line 122: Again with the yq. At the first sign of problems, perhaps we switch this to use `grep -v`.

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -173,6 +173,7 @@ prow-config:
 codecov-secret-mapping:
 	${CONVENTION_DIR}/codecov-secret-mapping ${RELEASE_CLONE}
 
+
```
2uasimojo, line 176: Random whitespace damage.

test/projects/file-generate/config/templates/csv-template.yaml
```diff
@@ -0,0 +1,45 @@
...
+  name: file-generate-1.0.0
+  namespace: placeholder
+  annotations:
+    categories: A list of comma separated categories that your operator falls under.
+    certified: "false"
+    description: Test operator
+    containerImage: quay.io/app-sre/file-generate:latest
+    createdAt: "2019-04-10T17:34:33Z"
+    support: Carl Brumm
+spec:
```
2uasimojo, line 13: Do we want this to have a `replaces` or not?

2uasimojo (approved):
There's plenty of work to be done here, but I feel like this is in good enough shape that we could land it and iterate.

/lgtm

but I know @jharrington22 wanted another look at this, so

/hold

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
bdematte: Objective was to avoid having logic exclusive to the common version which couldn't be supported for the 'hack' scripts. However, if we consider any new operator should use the common version, we could move it to the python script indeed.

test/projects/file-generate/config/templates/csv-template.yaml
bdematte: I would rather not have a replace as this is used to init a project without former version.

2uasimojo: Sure: we either include it, and delete it if necessary; or exclude it, and add it if necessary. I don't care either way, but the code and the yaml should all agree :)

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
2uasimojo: If you want to keep yq but be sure which one you're using, you could [use a containerized version](https://github.com/openshift/pagerduty-operator/blob/a64ef58978d1427e0956eea390599dab4352eef8/functions.mk#L22).

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
```diff
@@ -0,0 +1,148 @@
...
+if [ -z "$GIT_PATH" ] ; then 
+    GIT_PATH="https://app:@gitlab.cee.redhat.com/service/saas-${operator_name}-bundle.git"
+fi
+
+# Calculate previous version
+SAAS_OPERATOR_DIR="saas-${operator_name}-bundle"
+BUNDLE_DIR="$SAAS_OPERATOR_DIR/${operator_name}/"
+
+if [ "$diff_generate" = true ] ; then
+    OPERATOR_NEW_VERSION=$(ls "$BUNDLE_DIR" | sort -t . -k 3 -g | tail -n 1)
```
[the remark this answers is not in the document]
jharrington22: @bdematte doesn't look like this is addressed unless there is something I'm misunderstanding? Since you `rm` the  `$SAAS_OPERATOR_DIR` in the next else branch I guess you expect it to exist if `$diff_generate = true` ? We should validate that before this if statement as this would fail and the `rm` in the next branch would exit because `set -e` is set.

jharrington22 (commented):
Just two comments and a final piece which could be carded up. 

What happens with `$OPERATOR_PREV_VERSION` and `$OPERATOR_NEW_VERSION` version are the same? This could happen if `staging-catalog-publish` or `production-catalog-publish` fail and we rebuild the same job. The commit will exist in the SAAS repo and the old will equal new. 

This is a problem we have today so we don't need to fix it in this PR but at least create a card for it if its not handled here already (I don't think it is)

bdematte:
> Just two comments and a final piece which could be carded up.
> 
> What happens with `$OPERATOR_PREV_VERSION` and `$OPERATOR_NEW_VERSION` version are the same? This could happen if `staging-catalog-publish` or `production-catalog-publish` fail and we rebuild the same job. The commit will exist in the SAAS repo and the old will equal new.
> 
> This is a problem we have today so we don't need to fix it in this PR but at least create a card for it if its not handled here already (I don't think it is)

The idea was to just use the dedicated `make` targets but indeed, would need to be able to rebuild the CSV too as we do not commit it in the repo. 

For the 2 other points you mentioned, there are already fixed in the fup-MR I prepared (will submit it once this one is merged if fine to avoid 'monster reviews')

bdematte:
/unhold
as discussed with James

PR #103, opened 2020-12-01
Support containerized local builds

2uasimojo:
Add a utility, boilerplate/_lib/container-make, which can be used as a drop-in replacement for the real `make` to invoke the target(s) in the boilerplate container appropriate to the current level of boilerplate.

[OSD-5096](https://issues.redhat.com/browse/OSD-5096)

### DIFF
--- a/boilerplate/_lib/common.sh
+++ b/boilerplate/_lib/common.sh
@@ -99,5 +99,10 @@ fi
 # The namespace of the ImageStream by which prow will import the image.
 IMAGE_NAMESPACE=openshift
 IMAGE_NAME=boilerplate
+# LATEST_IMAGE_TAG may be set by `update`, in which case that's the
+# value we want to use.
+if [[ -z "$LATEST_IMAGE_TAG" ]]; then
+    LATEST_IMAGE_TAG=$(cat ${CONVENTION_ROOT}/_data/backing-image-tag)
+fi
 # The public image location
 IMAGE_PULL_PATH=quay.io/app-sre/$IMAGE_NAME:$LATEST_IMAGE_TAG
--- a/boilerplate/_lib/container-make
+++ b/boilerplate/_lib/container-make
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+source ${0%/*}/common.sh
+
+CONTAINER_ENGINE=$(command -v podman || command -v docker)
+[[ -n "$CONTAINER_ENGINE" ]] || err "Couldn't find a container engine. Are you already in a container?"
+
+# echo this regardless of BOILERPLATE_SET_X
+set -x
+$CONTAINER_ENGINE run --rm -v "$REPO_ROOT":"$REPO_ROOT" $IMAGE_PULL_PATH /bin/sh -c "cd $REPO_ROOT; make $@"
--- a/boilerplate/update
+++ b/boilerplate/update
@@ -186,10 +186,16 @@ for convention in $(scrubbed_conventions $CONFIG_FILE); do
   echo ""
 done
 
-# If all that went well, lay down the last-boilerplate-commit file,
-# which allows freeze-check to work.
-echo "Registering commit hash..."
+# If all that went well, record some metadata.
 mkdir -p ${CONVENTION_ROOT}/_data
+# - The last-boilerplate-commit file, which allows freeze-check to work.
+echo "Registering commit hash..."
 bp_commit=$(cd ${BP_CLONE} && git rev-parse HEAD)
 echo ${bp_commit} > ${CONVENTION_ROOT}/_data/last-boilerplate-commit
+
+# - The boilerplate backing image tag. This is used to run containerized
+#   local builds/tests.
+echo "Registering image tag..."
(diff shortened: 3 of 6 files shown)


2uasimojo:
Notes to reviewers:
- Design-wise, I considered doing [this kind of thing](https://github.com/openshift/cloud-ingress-operator/blob/500ad47b77d7e96c88f3ee60b9ac1fee61d22540/Makefile#L28-L34) instead, but ultimately rejected it because a) we would have to duplicate very similar logic in many targets; b) dependency chains would then spin up N containers instead of just one (which is not as accurate a simulation of CI as this; and might actually break too).
- I tried to figure out how to make the container run successfully as non-root, but gave up. I think to make it work we might need to copy the source dir to a fresh user-owned directory in the container, which would be really heavy and slow.
- It would be nice to make this, and the other executable utilities in `boilerplate/_lib`, easier to invoke. Perhaps by moving/symlinking them in a new `bin/` directory? But that's for a separate PR.

jharrington22 (commented):
@2uasimojo nice, concern is that the boilerplate image will become bloated with tools required to build operators etc. It may be nice to be able to override this image in the future with a build container specified from the consumer. 

@2uasimojo is this not trying to build docker containers in a docker container?

2uasimojo:
> concern is that the boilerplate image will become bloated with tools required to build operators etc.

Couple things here.
1. I'm confused. That's the whole point of the boilerplate image: to contain tools required to build (and test, validate, etc.) operators.
2. As for "bloated" -- the worst part of it right now is the golang-1.13 base, which is enormous. The stuff we're adding to that base is tiny.
3. This PR isn't changing the image at all. Maybe this discussion is for elsewhere?

> It may be nice to be able to override this image in the future with a build container specified from the consumer.

That would defeat the purpose of boilerplate, whose whole charter is to standardize the tools we use to build and validate operators. But if you wanted to use a bespoke container to do those things, nothing about this PR would be stopping you.

> @2uasimojo is this not trying to build docker containers in a docker container?

Well, if you ran a target that itself uses docker, like `container-make docker-build`. But the purpose of this wrapper is so you can run things like `container-make validate` and know that you're doing the same thing as prow will do.

jharrington22:
@2uasimojo my concern was around the fact that you can pass any make command to this, so a consumer could run `container-make docker-build` which would fail. A user could also try to build their whole application with this which might require more dependencies in the docker image etc.

Using this for all the validations pieces to emulate prow etc is great and its intended purpose. For the beginning is the intention that a user would run this command manually?

2uasimojo:
Sure, they can do all kinds of things that won't work. `container-make docker-build` would at least have the courtesy to break, hard, right away. What's worse -- and what we're trying to avoid -- is e.g. if they run an ordinary `make lint` with the wrong (or no) version of golangci-lint installed, get a false positive (or negative), react accordingly, and then push a PR that breaks in prow.

> Using this for all the validations pieces to emulate prow etc is great and its intended purpose. For the beginning is the intention that a user would run this command manually?

Yes. Not even just "the beginning" -- that's its entire purpose. (Can you think of some other way this could be used?)

jharrington22:
> Sure, they can do all kinds of things that won't work. `container-make docker-build` would at least have the courtesy to break, hard, right away. What's worse -- and what we're trying to avoid -- is e.g. if they run an ordinary `make lint` with the wrong (or no) version of golangci-lint installed, get a false positive (or negative), react accordingly, and then push a PR that breaks in prow.
> 
> > Using this for all the validations pieces to emulate prow etc is great and its intended purpose. For the beginning is the intention that a user would run this command manually?
> 
> Yes. Not even just "the beginning" -- that's its entire purpose. (Can you think of some other way this could be used?)

Make sense, I was thinking of a pre-commit hook that run the prow equivalent checks.

jharrington22 (approved):

2uasimojo:
> Make sense, I was thinking of a pre-commit hook that run the prow equivalent checks.

Ah, gotcha. Yeah, I don't see that being a significant savings over letting prow do the same. But running manually, you can iterate, just rerunning the one you're working on, etc.

Thanks for reviewing!

jharrington22:
> > Make sense, I was thinking of a pre-commit hook that run the prow equivalent checks.
> 
> Ah, gotcha. Yeah, I don't see that being a significant savings over letting prow do the same. But running manually, you can iterate, just rerunning the one you're working on, etc.
> 
> Thanks for reviewing!

It would be a significant time saving, especially things like lint.

2uasimojo:
> It would be a significant time saving, especially things like lint.

I think we're violently agreeing here, but just to be clear:
If we ran the prow checks via a precommit hook, even in parallel, that hook would take ~ the same amount of time to run locally as it would if I pushed and let the CI cluster do it. But I often commit multiple times while developing locally, and I definitely don't want all of these checks to run every time.
What saves me time is if I can run the checks, see that one of them (e.g. lint) fails, and then iterate on just that check.

Now, correct me if I'm wrong, but I don't think anything stops individuals from setting up precommit hooks with these checks if they wish?

PR #104, opened 2020-12-02
Add opm-build-push target

rporres:
* Opt-in target, we are not ready to force anyone to use this
* Adds a script to build catalog using opm
* Adds authentication to docker-push (breaking change)
* Depends on a local `hack/generate-bundle-contents.py` script

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
+++ b/boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
@@ -1,0 +1,380 @@
+#!/bin/bash
+
+set -euo pipefail
+
+## Global vars to modify behaviour of the script
+SET_X=${SET_X:-false}
+[[ "$SET_X" != "false" ]] && set -x
+
+DRY_RUN=${DRY_RUN:-false}
+DELETE_TEMP_DIR=${DELETE_TEMP_DIR:-true}
+NO_LOG=${NO_LOG:-false}
+OLM_BUNDLE_VERSIONS_REPO=${OLM_BUNDLE_VERSIONS_REPO:-gitlab.cee.redhat.com/service/saas-operator-versions.git}
+OLM_BUNDLE_VERSIONS_REPO_BRANCH=${OLM_BUNDLE_VERSIONS_REPO_BRANCH:-master}
+
+# Global vars
+REPO_ROOT=$(git rev-parse --show-toplevel)
+
+function log() {
+    local to_log=${1}
+    [[ "$NO_LOG" != "false" ]] && return 0
+    local msg
+    msg="$(date "+%Y-%m-%d %H:%M:%S")"
+    [[ "$DRY_RUN" != "false" ]] && msg="$msg -- DRY-RUN"
+    echo "$msg -- $to_log" 1>&2
+}
+
+function check_required_environment() {
+    local count=0
+    for var in OLM_BUNDLE_IMAGE \
+               OLM_CATALOG_IMAGE \
+               OPERATOR_IMAGE \
+               OPERATOR_IMAGE_TAG \
+               OPERATOR_VERSION \
+               OPERATOR_NAME \
+               CONTAINER_ENGINE \
+               CONTAINER_ENGINE_CONFIG_DIR \
+               CURRENT_COMMIT \
+               OLM_CHANNEL
+    do
+        if [ ! "${!var:-}" ]; then
+          log "$var is not set"
+          count=$((count + 1))
+        fi
+    done
+
+    [[ $count -eq 0 ]] && return 0 || return 1
+}
+
+function setup_temp_dir() {
+    local temp_dir
+    temp_dir=$(mktemp -d --suffix "-$(basename "$0")")
+    [[ "$DELETE_TEMP_DIR" == "true" ]] && trap 'rm -rf $temp_dir' EXIT
+
+    echo "$temp_dir"
+}
+
+function setup_local_executable() {
+    local executable=${1}
+    "$REPO_ROOT"/boilerplate/openshift/golang-osd-operator/ensure.sh "$executable" >&2
+    echo "$REPO_ROOT/.$executable/bin/$executable"
+}
+
+function check_bundle_contents_cmd() {
+    bundle_contents_cmd="$REPO_ROOT/hack/generate-operator-bundle-contents.py"
+    if [[ ! -x "$bundle_contents_cmd" ]]; then
+        log "$bundle_contents_cmd is either missing or non-executable"
+        return 1
+    fi
+}
+
+# Check we are running an opm supported container engine
+function check_opm_supported_container_engine() {
+    local image_builder=${1}
+    if [[ "$image_builder" != "docker" && "$image_builder" != "podman" ]]; then
+        # opm error messages are obscure. Let's make this clear
+        log "image_builder $image_builder is not one of docker or podman"
+        return 1
+    fi
+
+    return 0
+}
+
+# Check we will be able to properly authenticate ourselves against the registry
+function check_authenticated_registry_command() {
+    if [[ ! -f "$CONTAINER_ENGINE_CONFIG_DIR/config.json" ]]; then
+        log "$CONTAINER_ENGINE_CONFIG_DIR/config.json missing"
+        return 1
+    fi
+
+    return 0
+}
+
+function setup_authenticated_registry_command() {
+    echo "$CONTAINER_ENGINE --config=$CONTAINER_ENGINE_CONFIG_DIR"
+}
+
+function clone_olm_bundle_versions_repo() {
+    local saas_root_dir=${1}
+
+    local bundle_versions_repo_url
+    if [[ -n "${APP_SRE_BOT_PUSH_TOKEN:-}" ]]; then
+        log "Using APP_SRE_BOT_PUSH_TOKEN credentials to authenticate"
+        bundle_versions_repo_url="https://app:${APP_SRE_BOT_PUSH_TOKEN}@$OLM_BUNDLE_VERSIONS_REPO"
+    else
+        bundle_versions_repo_url="https://$OLM_BUNDLE_VERSIONS_REPO"
+    fi
+
+    log "Cloning $OLM_BUNDLE_VERSIONS_REPO into $saas_root_dir"
+    git clone --branch "$OLM_BUNDLE_VERSIONS_REPO_BRANCH" "$bundle_versions_repo_url" "$saas_root_dir"
+}
+
+function get_prev_operator_version() {
+    local bundle_versions_file=${1}
+
+    local return_message=""
+
+    # if the line contains SKIP it will not be included
+    local prev_operator_version=""
+    local prev_good_operator_version=""
+    local skip_versions=()
+    if [[ -s "$bundle_versions_file" ]]; then
+        log "$bundle_versions_file exists. We'll use to determine current version"
+
+        prev_operator_version=$(tail -n 1 "$bundle_versions_file" | awk '{print $1}')
+
+        # we traverse the bundle versions file backwards
+        # we cannot use pipes here or we would lose the inner variables changes
+        local version
+        while read -r line; do
+            if [[ "$line" == *SKIP* ]]; then
+                version=$(echo "$line" | awk '{print $1}')
+                skip_versions+=("$version")
+            else
+                prev_good_operator_version="$line"
+                break
+            fi
+        done < <(sort -r -t . -k 3 -g "$bundle_versions_file")
+
+        if [[ -z "$prev_good_operator_version" ]]; then
+            # This means that we have skipped all the available versions. In this case we're going to use the last
+            # SKIP version as the prev_good_operator_version to have something to feed replaces in the CSV
+            log "No unskipped version in $bundle_versions_file. We'll use the last skipped one: ${skip_versions[0]}"
+            prev_good_operator_version="${skip_versions[0]}"
+        fi
+
+        log "Previous operator version is $prev_operator_version"
+        log "Previous good operator version is $prev_good_operator_version"
+        return_message="$prev_operator_version $prev_good_operator_version"
+
+        if [[ ${#skip_versions[@]} -gt 0 ]]; then
+            log "We will be skipping: ${skip_versions[*]}"
+            return_message="$return_message ${skip_versions[*]}"
+        fi
+    else
+        log "No $bundle_versions_file exist or it is empty. This is the first time the operator is built"
+        if [[ ! -d "$(dirname "$bundle_versions_file")" ]]; then
+            log "Operator directory doesn't exist in versions repository. Exiting"
+            exit 1
+        fi
+    fi
+
+    echo "$return_message"
+
+}
+
+function build_opm_bundle() {
+    local bundle_contents_cmd=${1}
+    local opm_local_executable=${2}
+    local image_builder=${3}
+    local prev_good_operator_version=${4}
+    # shellcheck disable=SC2206
+    local skip_versions=(${@:5})
+
+    local bundle_temp_dir
+    bundle_temp_dir=$(mktemp -d -p "$temp_dir" bundle.XXXX)
+    local generate_csv_template_args=""
+    [[ -n "$prev_good_operator_version" ]] && generate_csv_template_args="--replaces $prev_good_operator_version"
+    if [[ ${#skip_versions[@]} -gt 0 ]]; then
+        for version in "${skip_versions[@]}"; do
+            generate_csv_template_args="$generate_csv_template_args --skip $version"
+        done
+    fi
+
+    local manifests_temp_dir
+    manifests_temp_dir=$(mktemp -d -p "$bundle_temp_dir" manifests.XXXX)
+    # shellcheck disable=SC2086
+    $bundle_contents_cmd --name "$OPERATOR_NAME" \
+                         --current-version "$OPERATOR_VERSION" \
+                         --image "$OPERATOR_IMAGE" \
+                         --image-tag "$OPERATOR_IMAGE_TAG" \
+                         --output-dir "$manifests_temp_dir" \
+                         $generate_csv_template_args
+
+    # opm won't get anything locally, so we need to push the bundle even in dry run mode
+    # we will use a different tag to make sure those leftovers are clearly recognized
+    # TODO: remove this tag if we're in dry-run mode in the cleanup trap script
+    [[ "$DRY_RUN" == "false" ]] || bundle_image_current_commit="${bundle_image_current_commit}-dryrun"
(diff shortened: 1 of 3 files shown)


rporres:
A few notes:

* Missing documentation. I will create once we like the approach
* local `hack/generate-bundle-contents.py` script should be moved to boilerplate. It should follow the conventions from https://github.com/openshift/boilerplate/pull/101
* Previous `docker-push` was unauthenticated. I made it depend from login and build. This is a breaking change
* Tested in deployment-validation-operator

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -80,13 +89,19 @@ docker-build: isclean
...
-	${CONTAINER_ENGINE} push $(OPERATOR_IMAGE_URI)
-	${CONTAINER_ENGINE} push $(OPERATOR_IMAGE_URI_LATEST)
+docker-push: docker-login docker-build
+	${CONTAINER_ENGINE} --config=${CONTAINER_ENGINE_CONFIG_DIR} push ${OPERATOR_IMAGE_URI}
+	${CONTAINER_ENGINE} --config=${CONTAINER_ENGINE_CONFIG_DIR} push ${OPERATOR_IMAGE_URI_LATEST}
 
 .PHONY: push
 push: docker-push
 
+.PHONY: docker-login
```
2uasimojo, line 99: Is there a way to check whether already authenticated? If so, it would be preferable to do that, and only require/use the cred vars if necessary.

2uasimojo (commented):
Having already reviewed this code in dvo itself, I'll just say I'm on board with the overall concepts.

We discussed this in TΩ this morning and want to

/assign @bdematte 

to dig into how the various pieces should be split up. We want this and #101 to evolve together, and use common chunks where approriate.

boilerplate/openshift/golang-osd-operator/standard.mk
rporres: The presence of docker configuration doesn't guarantee that it has a working set of credentials, so I don't see how this could be done easily.

On second thoughts, docker pushes must be protected as much as possible. They are one of the most delicate parts of the system and we know the consequences of bad pushes to both the catalog or the operator image... I don't think it is a bad thing that creds are always requested.

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -30,6 +30,15 @@ OPERATOR_IMAGE_URI=${IMG}
 OPERATOR_IMAGE_URI_LATEST=$(IMAGE_REGISTRY)/$(IMAGE_REPOSITORY)/$(IMAGE_NAME):latest
 OPERATOR_DOCKERFILE ?=build/Dockerfile
 
+OLM_BUNDLE_IMAGE = $(OPERATOR_IMAGE)-bundle
+OLM_CATALOG_IMAGE = $(OPERATOR_IMAGE)-catalog
+OLM_CHANNEL ?= alpha
+OLM_BUNDLE_VERSIONS_REPO ?= gitlab.cee.redhat.com/service/saas-deployment-validation-operator-bundle.git
```
bdematte, line 36: Shouldn't it be gitlab.cee.redhat.com/service/saas-operator-versions.git ?

bdematte (changes requested):

boilerplate/openshift/golang-osd-operator/standard.mk
2uasimojo: +1 https://github.com/app-sre/deployment-validation-operator/blob/master/Makefile#L1

2uasimojo: Though thinking about it, why not just hardcode it in the script itself, since it's going to be the same for every consumer? That'll save us a parameter/envvar :)

rporres: Good catch. Yes, I will change it. I would leave as parametrizable, I like the possibility of overriding it, e.g in testing scenarios.

rporres: Fixed in ea1c9f7

rporres:
What's missing here? I understand `hack/generate-bundle-contents.py` should be generalized but I don't know if we want to wait until #101 is completed to move on with this PR.

bdematte:
> What's missing here? I understand `hack/generate-bundle-contents.py` should be generalized but I don't know if we want to wait until #101 is completed to move on with this PR.

For me, we are missing documentation and testing for the PR to be mergeable. 
I don't think we should include more changes in this one to start - we can go with the rest (spliting it, generalizing the generate-bundle, and so on) in dedicated PRs.

rrati:
@bdematte We've done testing with DVO for these changes.  What are you looking for re: testing and docs?

rporres:
The documentation is missing, as stated before I didn't want to write it before we agreed with the approach.

What kind of testing do you want to see here?

rporres:
For example, the catalog created via `opm` has been thoroughly tested: https://issues.redhat.com/browse/APPSRE-2741

End to end tests for this would be very much desirable, but not easily achievable since they involve many moving parts.

bdematte:
We need to have integrated CI testing for it as this will be evolving in the coming weeks/month. 
PR#101 contains a first version of the testing for the 'old way' and you can use part of it for the testing of this I believe.

rporres:
A few questions/comments about tests to be able to come with a valid strategy:

1. Where would this tests run? In a VM or in a docker container. This is important because the script uses extensively docker (`opm` uses docker internally, we run the catalog to verify it returns the proper version) and that would mean running docker in docker, which is always a challenge to say the least
2. The code would need to be split in parts either via functions or makefile targets. Makefile targets would be in line with the tests outlined in #101 but since targets would need to be wrapped in another script (they share lots of variables) that would mean that a script that is called from a Makefile target would call another targets of the same Makefile instead of using the common dependency approach. That looks hacky...
3. Bundle generation can not be tested as it is outside the code
4. Which is the test runner? Can something as bats be used?

bdematte:
CI is running those tests through docker indeed. 
Without splitting it, we should at least be able to test that we can create create CSV for staging/production, and that the SKIP mechanism reflect properly (in gitlab and in the CSV). There are other cases that would be good to add eventually, but I think this would be the bare minimum and wouldn't require a split. 
`driver` is pretty simple and is running any executable files detected in the `test/case`

jharrington22:
@rporres I review this and it looks good. What @bdematte is asking for is some tests so we can ensure iterating on this moving forward is safe, but given that it would require running docker in docker its going to be difficult.

Looking at the script I think having the following would be enough and shouldn't be too much work
- validate that the `generate-bundle-contents.py` is generating what we expect each time
- validating the parsing of `bundle_versions_file`
- validating the catalog build (looks like thats already added in the script) - is a dry run enough?

I'm ok with the documentation coming after this PR since things are likely to change pretty quickly and there isn't adoption from SRE operators just yet, we'll likely have to write something help people migrate.

@bdematte @rporres I don't want to hold up this PR, I'd like it to merge before code freeze so I'm OK If the tests are very simple static validations to begin with.

rporres:
@jharrington22 I've had a meeting with @bdematte to come with a strategy to have this merged before code freeze:

> validate that the generate-bundle-contents.py is generating what we expect each time

not feasible. That's a local script in `dvo` repo. I can move it into a target that has to be overriden in the boilerplate consumer to make it more flexible

> validating the parsing of bundle_versions_file

I will extract that logic into a separate script so that it can be tested

> validating the catalog build (looks like thats already added in the script) - is a dry run enough?

I don't see how to do this as validating the catalog involves running the container... Dry run mode has proven to be very helpful to me while developing, I think it should be enough

> @bdematte @rporres I don't want to hold up this PR, I'd like it to merge before code freeze so I'm OK If the tests are very simple static validations to begin with.

I think we have time before Wednesday 8pm to add some tests and review this. Let's try to do it :)

jharrington22:
@rporres sounds great! Thank you.

rporres:
In order to add tests I had to refactor the original code into pieces. Done in c78b5a7
Tests are in 47dfad7

test/case/convention/openshift/golang-osd-operator/05-build-olm-catalog
```diff
@@ -0,0 +1,210 @@
...
+export OLM_CATALOG_IMAGE=quay.io/rporresm/deployment-validation-operator-catalog
+export CONTAINER_ENGINE=/usr/local/bin/docker
+export CONTAINER_ENGINE_CONFIG_DIR=.docker
+export CURRENT_COMMIT=860gta5q
+export COMMIT_NUMBER=5
+export OPERATOR_VERSION=0.1.5-860gta5q
+export OPERATOR_NAME=deployment-validation-operator
+export OPERATOR_IMAGE=quay.io/rporresm/deployment-validation-operator
+export OPERATOR_IMAGE_TAG=not-important-at-the-moment
+export OLM_BUNDLE_VERSIONS_REPO=gitlab.cee.redhat.com/service/saas-deployment-validation-operator-bundle.git
```
bdematte, line 17: Shouldn't it be "gitlab.cee.redhat.com/service/saas-operator-versions.git" also here ?

rporres: we don't really care... it is not used in the unit tests (and should not be used 😄 ). I will change it to be `not-important-at-the-moment` to further clarify it

rporres: fixed in cc6b3f6

boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
```diff
@@ -0,0 +1,379 @@
...
+    local temp_dir
+    temp_dir=$(mktemp -d --suffix "-$(basename "$0")")
+    [[ "$DELETE_TEMP_DIR" == "true" ]] && trap 'rm -rf $temp_dir' EXIT
+
+    echo "$temp_dir"
+}
+
+function setup_local_executable() {
+    local executable="${1}"
+    "$REPO_ROOT"/boilerplate/openshift/golang-osd-operator/ensure.sh opm
```
2uasimojo, line 58: 1. Use the param.
2. Do you need to make sure this command doesn't print anything to stdout, since you're relying on the output from the function?
```suggestion
    # Redirect to stderr to make sure this output doesn't interfere with stdout expected from the function
    "$REPO_ROOT"/boilerplate/openshift/golang-osd-operator/ensure.sh ${executable} >&2
```

boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
```diff
@@ -0,0 +1,379 @@
...
+               OLM_CATALOG_IMAGE \
+               OPERATOR_IMAGE \
+               OPERATOR_IMAGE_TAG \
+               OPERATOR_VERSION \
+               OPERATOR_NAME \
+               CONTAINER_ENGINE \
+               CONTAINER_ENGINE_CONFIG_DIR \
+               CURRENT_COMMIT \
+               OLM_CHANNEL \
+               OLM_BUNDLE_VERSIONS_REPO
```
2uasimojo, line 37: Since the bundle versions repo is so tightly bound to other variables that aren't "parameters" to this script (e.g. the path to the versions file), suggest simply hardcoding it and getting rid of this variable. (IMO the more we can reduce the number of env vars we're relying on, the better.)

2uasimojo (commented):
Partial review, more coming.

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -80,13 +89,19 @@ docker-build: isclean
...
+docker-push: docker-login docker-build
+	${CONTAINER_ENGINE} --config=${CONTAINER_ENGINE_CONFIG_DIR} push ${OPERATOR_IMAGE_URI}
+	${CONTAINER_ENGINE} --config=${CONTAINER_ENGINE_CONFIG_DIR} push ${OPERATOR_IMAGE_URI_LATEST}
 
 .PHONY: push
 push: docker-push
 
+.PHONY: docker-login
+docker-login:
+	@test "${REGISTRY_USER}" != "" || test "${REGISTRY_TOKEN}" != "" || (echo "REGISTRY_USER and REGISTRY_TOKEN must be defined" && exit 1)
```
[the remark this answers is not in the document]
rporres: yes we want... let me take a look

boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
rporres: 1. Yes. That's a bug. It was basically working because I had already everything installed
2. Yes. I wonder how that was working in the first place. I'll fix it

boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
rporres: I was leaving this as a parameter because it's helpful to be able to test against a different git repository if needed. But I can remove it and make it overridable if needed to remove it from the Makefiles as it is just testing convenient.

boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
rporres: 2. `ensure.sh` throws everything to stderr, that's why it was working. I'll add your suggestion as it is safer

rporres: fixed in 3424552

boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
rporres: Fixed in 1bd829f

2uasimojo (commented):
Partial review part deux.

boilerplate/openshift/golang-osd-operator/standard.mk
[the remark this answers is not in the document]
2uasimojo: :+1:

boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
```diff
@@ -0,0 +1,379 @@
...
+# Global vars
+REPO_ROOT=$(git rev-parse --show-toplevel)
+
+## Global vars to modify behaviour of the script
+DRY_RUN=${DRY_RUN:-false}
+DELETE_TEMP_DIR=${DELETE_TEMP_DIR:-true}
+BRANCH=${BRANCH:-master}
+SET_X=${SET_X:-false}
+NO_LOG=${NO_LOG:-false}
+[[ "$SET_X" != "false" ]] && set -x
```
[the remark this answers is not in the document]
rporres: fixed in 3e8ecf1

boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
```diff
@@ -0,0 +1,379 @@
...
+
+set -euo pipefail
+
+# Global vars
+REPO_ROOT=$(git rev-parse --show-toplevel)
+
+## Global vars to modify behaviour of the script
+DRY_RUN=${DRY_RUN:-false}
+DELETE_TEMP_DIR=${DELETE_TEMP_DIR:-true}
+BRANCH=${BRANCH:-master}
```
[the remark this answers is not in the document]
rporres: `OLM_BUNDLE_VERSIONS_REPO_BRANCH` fixed in 6a583f7

boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
```diff
@@ -0,0 +1,379 @@
...
+        fi
+    else
+        log "No $bundle_versions_file exist or it is empty. This is the first time the operator is built"
+        if [[ ! -d "$(dirname "$bundle_versions_file")" ]]; then
+            log "Operator directory doesn't exist in versions repository. Exiting"
+            exit 1
+        fi
+    fi
+
+    echo "$return_message"
```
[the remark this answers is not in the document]
rporres: Yeah, that could be potentially dangerous... On the other hand, that's THE function that we have thoroughly tested 😂 

I'll take a look into that after we have merged this, to see if I can come with something in the lines you're suggesting

boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
```diff
@@ -0,0 +1,379 @@
...
+
+function get_prev_operator_version() {
+    local bundle_versions_file=${1}
+
+    local return_message=""
+
+    # if the line contains SKIP it will not be included
+    local prev_operator_version=""
+    local prev_good_operator_version=""
+    skip_versions=()
```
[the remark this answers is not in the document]
rporres: fixed in de4b43c

boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
```diff
@@ -0,0 +1,379 @@
...
+## Global vars to modify behaviour of the script
+DRY_RUN=${DRY_RUN:-false}
+DELETE_TEMP_DIR=${DELETE_TEMP_DIR:-true}
+BRANCH=${BRANCH:-master}
+SET_X=${SET_X:-false}
+NO_LOG=${NO_LOG:-false}
+[[ "$SET_X" != "false" ]] && set -x
+
+function log() {
+    local to_log=${1}
```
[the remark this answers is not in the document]
rporres: No need for that. It's only needed when you interpolate variables, not when you define them

boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
```diff
@@ -0,0 +1,379 @@
...
+
+        if [[ ${#skip_versions[@]} -gt 0 ]]; then
+            log "We will be skipping: ${skip_versions[*]}"
+            return_message="$return_message ${skip_versions[*]}"
+        fi
+    else
+        log "No $bundle_versions_file exist or it is empty. This is the first time the operator is built"
+        if [[ ! -d "$(dirname "$bundle_versions_file")" ]]; then
+            log "Operator directory doesn't exist in versions repository. Exiting"
+            exit 1
```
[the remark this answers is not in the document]
rporres: empty message has a different meaning... It means no previous version ever existed. I will take care of this after this is merged as I want to add a test case for it. I agree that `exit`s should only be in `main`

rporres:
Squashed all the commits into one. I have left a couple of things to be done afterwards, as I'm running out of time today. We can totally live with them while I fix them.

2uasimojo (approved):
I haven't finished reviewing, but James and Ben have been through it, and since it's not in the way of anything, we've agreed to land it now and iterate as needed.

/lgtm
/approve

PR #105, opened 2020-12-07
image-v0.4.0

2uasimojo:
This PR puts together `image-v0.4.0` with the following changes:
- Don't set up gvm anymore. We don't need it.
- Include mockgen and openapi-gen.

See individual commit messages for details.

### DIFF
--- a/config/build.sh
+++ b/config/build.sh
@@ -59,35 +59,6 @@ mv yq /usr/local/bin
 ##################
 python3 -m pip install PyYAML==5.3.1
 
-##################
-# golang (via gvm)
-##################
-GO_VERSIONS="go1.13.15 go1.14.10"
-
-GVM_VERSION=1.0.22
-GVM_SHA256SUM="72123889c8ef55f7b745038dc8cf556c1aece982408966fa5ff612ce6a97bae7"
-GVM_LOCATION=https://raw.githubusercontent.com/moovweb/gvm/${GVM_VERSION}/binscripts/gvm-installer
-
-curl -L -o gvm-installer $GVM_LOCATION
-echo ${GVM_SHA256SUM} gvm-installer | sha256sum -c
-chmod ugo+x gvm-installer
-./gvm-installer
-rm -f gvm-installer
-source /root/.gvm/scripts/gvm
-
-GVM_DEPS="bison"
-yum -y install ${GVM_DEPS}
-
-for GO_VERSION in $GO_VERSIONS; do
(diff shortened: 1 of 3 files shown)


2uasimojo:
/hold

We're seeing permission issues in the go installation. Not really sure what to do about that. Will dig in tomorrow.

jharrington22:
> /hold
> 
> We're seeing permission issues in the go installation. Not really sure what to do about that. Will dig in tomorrow.

Eh, how does this generate permission errors?

2uasimojo:
> Eh, how does this generate permission errors?

I don't know yet. What I do know is that the pr-check failures are preceded by permission errors under the GOPATH. At this point I'm speculating it's one of several things:
- With gvm, we would have been building under a GOPATH that had been made writable explicitly. This PR rips out gvm, so we're back to using the stock go installation from the base image.
- That doesn't explain why image-v0.2.x was working, because it was pre-gvm. But maybe we didn't have the failing tests in place yet. Or maybe the errors were there, but didn't cause failures, so we didn't notice them.
- Could be a temporary glitch in the backing image. Remember, we're still building this with whatever's behind golang-1.13 at the time the job runs.
- The permission errors could be unrelated, and something else is causing the failures.

I'm going to look into it now. So

/retest

to get a new CI env.

2uasimojo:
/test pr-check

Missed my namespace again.

config/build.sh
```diff
@@ -42,6 +42,35 @@ for OPERATOR_SDK_VERSION in "${!OSDK_VERSION_HASHES[@]}"; do
...
+###############
+# Get rid of -mod=vendor
+unset GOFLAGS
+# No, really, we want to use modules
+export GO111MODULE=on
+
+#############
+# openapi-gen
+#############
+go get k8s.io/code-generator/cmd/openapi-gen@v0.19.4
```
dustman9000, line 56: Can we define all of these various versions into variables to the top of the file? It will make maintaining this much easier in the future.

config/build.sh
```diff
@@ -42,6 +42,35 @@ for OPERATOR_SDK_VERSION in "${!OSDK_VERSION_HASHES[@]}"; do
...
+
+#########
+# mockgen
+#########
+go get github.com/golang/mock/mockgen@v1.4.4
```
dustman9000, line 61: See comment above regarding defining a variable.

config/build.sh
2uasimojo: Done.

PR #109, opened 2021-01-08
pr-check: Color test output

2uasimojo:
...so it can be visually distinguished from the output of the commands run by the tests. This will make debugging easier.

### DIFF
--- a/test/lib.sh
+++ b/test/lib.sh
@@ -17,9 +17,29 @@ export NEXUS_MK=boilerplate/generated-includes.mk
 
 _BP_TEST_TEMP_DIRS=
 
+# ANSI colors
+RED='\033[0;31m'
+ORANGE='\033[0;33m'
+GREEN='\033[0;32m'
+RESET='\033[0m'
+
+colorprint() {
+  color=$1
+  shift
+  /bin/echo -e "${color}$@${RESET}"
+}
+
 err() {
-  echo "==ERROR== $@" >&2
+  colorprint $RED "==ERROR== $@" >&2
   exit 1
+}
+
+# Override "echo" so test output is visually distinguished.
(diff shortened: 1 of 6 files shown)


2uasimojo:
Here's a snippet of what the logs look like with this change (got this from `oc logs` from a [CI run](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_boilerplate/109/pull-ci-openshift-boilerplate-master-pr-check/1347663277135499264)):
Screenshot from 2021-01-08 16-26-27

test/lib.sh
```diff
@@ -18,10 +18,18 @@ export NEXUS_MK=boilerplate/generated-includes.mk
 _BP_TEST_TEMP_DIRS=
 
 err() {
-  echo "==ERROR== $@" >&2
+  echo -e "\e[31m==ERROR== $@\e[0m" >&2
```
dustman9000, line 21: I'd suggest defining these ANSI colors to a variable so we know what color this is.

test/lib.sh
```diff
@@ -18,10 +18,18 @@ export NEXUS_MK=boilerplate/generated-includes.mk
...
   exit 1
 }
 
+# Override "echo" so test output is visually distinguished.
+# NOTE: This means when you need to echo something raw (e.g. when
+# constructing a file, or as "output" from a function), you need to use
+# /bin/echo explicitly.
+echo() {
+  /bin/echo -e "\e[33m$@\e[0m"
```
dustman9000, line 30: Same here, define YELLOW and RESET ...

test/lib.sh
2uasimojo: Yeah, there's actually a way to use some kind of color name here instead. I'll look into it.

PR #111, opened 2021-01-14
openshift/golang-osd-operator: openapi-gen

2uasimojo:
Make consumers produce openapi by adding a target called `openapi-generate` which runs `openapi-gen` if necessary. Put it in the chain for `generate`.


arjunrn:
Is it possible to add a test for this target. Or is it overkill?

2uasimojo:
> Is it possible to add a test for this target. Or is it overkill?

It's already tested as part of 01-generated-files-checker. I added a condition in there to explicitly check the existence of the file generated by the new directive.

arjunrn:
/lgtm
/hold
Holding for other reviewers

PR #113, opened 2021-01-25
Improvements to container-make

2uasimojo:
- Run the container as self. Previously it was running as root, which had some unintended consequences, including leaving root-owned files in your repo.
- If the target fails, stay in the container so you can debug.
- Prints a usage statement if invoked with -h/--help.

### DIFF
--- a/boilerplate/_lib/container-make
+++ b/boilerplate/_lib/container-make
@@ -1,10 +1,36 @@
 #!/bin/bash
 
+if [[ "$1" == "-h"* ]] || [[ "$1" == "--h"* ]]; then
+  echo "Usage: $0 {arguments to the real 'make'}"
+  echo "Runs 'make' in the boilerplate backing container."
+  echo "If the command fails, starts a shell in the container so you can debug."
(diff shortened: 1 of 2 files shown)


boilerplate/_lib/container-make
```diff
@@ -1,10 +1,36 @@
...
+
+# Now run our `make` command in it with the right UID and working directory
+args="exec -it -u $(id -u):0 -w $REPO_ROOT $container_id"
+banner "Running: make $@"
+$CONTAINER_ENGINE $args make "$@"
+rc=$?
+
+# If it failed, drop into the container in a shell
+if [[ $rc -ne 0 ]]; then
+  banner "Oh noes, the command failed! Dropping you into the container so you can debug. Just 'exit' when you're done."
```
dustman9000, line 30:
```suggestion
  banner "Docker command failed! Dropping into shell for further debugging."
```

2uasimojo: Spoilsport :P

PR #114, opened 2021-01-28
Add missing push in git operations

rporres:
This was left behind in the refactoring made in #104

I've tested it via local changes in deployment-validation-operator https://github.com/app-sre/deployment-validation-operator/pull/82

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
+++ b/boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
@@ -286,6 +286,9 @@ function add_current_version_to_bundle_versions_file() {
 
     git commit -m "$message"
 
+    log "Pushing the repository changes to $OLM_BUNDLE_VERSIONS_REPO into $OLM_BUNDLE_VERSIONS_REPO_BRANCH branch"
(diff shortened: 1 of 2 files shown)


boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
```diff
@@ -286,6 +286,9 @@ function add_current_version_to_bundle_versions_file() {
 
     git commit -m "$message"
```
dustman9000, line 287: Seems like this should be wrapped with a `[[ "$DRY_RUN" == "false" ]]` as well?

rporres: No need to, because that is done locally in a temporal folder... I left it there so that you people can inspect how things would be pushed

rporres:
I realized that a test case was going to have a non-desired output

rporres:
@dustman9000 @2uasimojo Can you take a look into this?

boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
dustman9000: Sounds good, thanks!

PR #115, opened 2021-02-03
golang-osd-operator: Cap concurrency for codecov

2uasimojo:
Some consumers were regularly having their coverage CI pods OOM-killed. Experimenting with the prow configuration, we found we needed to set the memory limits unreasonably high to make it work.

By default, cover tool runs with a number of threads equal to the detected number of CPUs. In the CI environment at this time, that
appears to be 16. With this commit, we cap it at 4, which brings the memory usage down to a manageable level without unduly affecting the runtime of the job.

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/codecov.sh
+++ b/boilerplate/openshift/golang-osd-operator/codecov.sh
@@ -9,7 +9,12 @@ CI_SERVER_URL=https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test
 COVER_PROFILE=${COVER_PROFILE:-coverage.out}
 JOB_TYPE=${JOB_TYPE:-"local"}
 
-make -C "${REPO_ROOT}" go-test TESTOPTS="-coverprofile=${COVER_PROFILE}.tmp -covermode=atomic -coverpkg=./..."
+# Default concurrency to four threads. By default it's the number of procs,
+# which seems to be 16 in the CI env. Some consumers' coverage jobs were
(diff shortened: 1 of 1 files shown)


boilerplate/openshift/golang-osd-operator/codecov.sh
```diff
@@ -9,7 +9,11 @@ CI_SERVER_URL=https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test
 COVER_PROFILE=${COVER_PROFILE:-coverage.out}
 JOB_TYPE=${JOB_TYPE:-"local"}
 
-make -C "${REPO_ROOT}" go-test TESTOPTS="-coverprofile=${COVER_PROFILE}.tmp -covermode=atomic -coverpkg=./..."
+# `-p 4`: Set concurrency to four threads. By default it's the number of
+# procs, which seems to be 16 in the CI env. Some consumers' coverage
+# jobs were # regularly getting OOM-killed; so do this rather than boost
+# the pod resources unreasonably.
+make -C "${REPO_ROOT}" go-test TESTOPTS="-coverprofile=${COVER_PROFILE}.tmp -covermode=atomic -coverpkg=./... -p 4"
```
dustman9000, line 16: Can we make this a variable? Maybe COV_THREAD_COUNT?

2uasimojo: Good idea. Done.

jharrington22 (approved):
/lgtm

PR #116, opened 2021-02-09
Enable RMO onboarding

NautiluX:
* Allow configuration of MAINPACKAGE
* Allow configuration of TESTTARGETS

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/standard.mk
+++ b/boilerplate/openshift/golang-osd-operator/standard.mk
@@ -39,7 +39,7 @@ REGISTRY_TOKEN ?=
 CONTAINER_ENGINE_CONFIG_DIR = .docker
 
 BINFILE=build/_output/bin/$(OPERATOR_NAME)
-MAINPACKAGE=./cmd/manager
+MAINPACKAGE ?= ./cmd/manager
 
 GOOS?=$(shell go env GOOS)
 GOARCH?=$(shell go env GOARCH)
@@ -57,7 +57,9 @@ GOLANGCI_LINT_CACHE ?= /tmp/golangci-cache
 
 GOLANGCI_OPTIONAL_CONFIG ?=
 
+ifeq ($(origin TESTTARGETS), undefined)
 TESTTARGETS := $(shell ${GOENV} go list -e ./... | egrep -v "/(vendor)/")
+endif
 # ex, -v
 TESTOPTS :=
 


boilerplate/test/test-base-convention/standard.mk
```diff
@@ -37,7 +37,7 @@ GOENV=GOOS=${GOOS} GOARCH=${GOARCH} CGO_ENABLED=0 GOFLAGS=
 
 GOBUILDFLAGS=-gcflags="all=-trimpath=${GOPATH}" -asmflags="all=-trimpath=${GOPATH}"
 
-TESTTARGETS := $(shell ${GOENV} go list -e ./... | egrep -v "/(vendor)/")
+TESTTARGETS := $(shell ${GOENV} go list -e ./... | egrep -v "/(vendor)/" | grep -v /int)
```
2uasimojo, line 40: Couple things here:
- I'd like to get a bit more precise with the regex, as this will match any package anywhere whose name starts with `int`. Something like `egrep -v '/int(/.*)?$'` could work ("/int" followed by an optional subdirectory structure, then explicit EOL). 
- Can we please add a comment explaining why we're excluding this? I feel like `/vendor/` is fairly obvious, but this isn't.

All that said, I'm not loving that we're reserving "int", which seems like it could show up in any number of packages we legitimately want to test. It would be tough to notice that we're missing testing it, and then it would be awkward to resolve, especially if time had passed and we had already published e.g. `pkg/apis/int/v1`. If it were narrower, like `test/integration` (`/test/integration(/.*)?$`), I would feel more comfortable with it. (Obviously that would entail RMO re-pathing these tests. I would think that's not a show-stopper at this point.)

Another strategy is to override `TESTTARGETS` in your specific project. As written, you could simply assign it after the boilerplate `include`. Or we could change `:=` to `?=` here and the consumer could override it before or after. (The latter is probably better, in case we later add some other variable that derives from `TESTTARGETS` at the top level.)

Regardless, whatever we decide to do to boilerplate's make vars/rules will need to be done in boilerplate/openshift/golang_osd_operator/standard.mk. This file is only used for the internal test suites.

2uasimojo (changes requested):

boilerplate/test/test-base-convention/standard.mk
NautiluX: whoops selected the wrong file when I reapplied my changes. Yes, a way to override it would be a nicer I think. Will change the PR.

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -57,7 +57,7 @@ GOLANGCI_LINT_CACHE ?= /tmp/golangci-cache
 
 GOLANGCI_OPTIONAL_CONFIG ?=
 
-TESTTARGETS := $(shell ${GOENV} go list -e ./... | egrep -v "/(vendor)/")
+TESTTARGETS ?= $(shell ${GOENV} go list -e ./... | egrep -v "/(vendor)/")
```
2uasimojo, line 60: It turns out that the CI failure is a result of this variable having changed from "simply expanded" to "recursively expanded" [1]. I don't fully grok it, but I believe it has something to do with the expansion of `${GOENV}` needing to happen only when `TESTTARGETS` is defined, rather than when the `go-generate` target is evaluated. Because evaluating `${GOENV}` triggers a stealth `go mod tidy`?

Anyway, to my understanding, there is no "simply expanded" form of `?=`, so we have to work around it with something like:

```suggestion
ifeq ($(origin TESTTARGETS), undefined)
TESTTARGETS := $(shell ${GOENV} go list -e ./... | egrep -v "/(vendor)/")
endif
```

(See [2] for the equivalence of `?` and this `ifeq` business.)

I tested the above locally and it fixes the issue.

Arguably, we should do a deeper analysis of the relative benefits of simply vs recursively defining `${GOENV}` itself, but that's for another time.

[1] https://www.gnu.org/software/make/manual/html_node/Flavors.html#Flavors
[2] https://www.gnu.org/software/make/manual/html_node/Setting.html#Setting

2uasimojo:
I’m ready to merge this, but can we squash to reduce the noise in the commit history? (I don’t care about attribution if that’s why you separated.)

2uasimojo (approved):

boilerplate/openshift/golang-osd-operator/standard.mk
2uasimojo, line 60: oops, this looks like it went back to the original version.

2uasimojo:
/lgtm

Thanks @NautiluX

PR #118, opened 2021-03-15
Unbound error fix for the new operators

slaskawi:
This Pull Request fixes an error that appears when introducing a new Operator. 

In this case, there are no previous versions and the `get_prev_operator_version` function emits an empty string.  This string fails later on when trying to extract the first or the second item for the `versions` array.

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
+++ b/boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
@@ -319,6 +319,14 @@ function main() {
     local versions
     # shellcheck disable=SC2207
     versions=($(get_prev_operator_version "$bundle_versions_file"))
+    # This condition is triggered when an operator is built for the first time. In such case the
+    # get_prev_operator_version returns an empty string and causes undefined variables failures
+    # in a few lines below.
+    if [ -z ${versions+x} ]
+    then
+        versions[0]=""
+        versions[1]=""
+    fi
     local prev_operator_version="${versions[0]}"
     local prev_good_operator_version="${versions[1]}"
     local skip_versions=("${versions[@]:2}")


boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
```diff
@@ -319,6 +319,11 @@ function main() {
     local versions
     # shellcheck disable=SC2207
     versions=($(get_prev_operator_version "$bundle_versions_file"))
+    if [ -z ${versions+x} ]
+    then
+        versions[0]=""
+        versions[1]=""
```
2uasimojo, line 325: or
```suggestion
        versions=("" "")
```
but seriously, don't change this :)

boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
```diff
@@ -319,6 +319,11 @@ function main() {
     local versions
     # shellcheck disable=SC2207
     versions=($(get_prev_operator_version "$bundle_versions_file"))
+    if [ -z ${versions+x} ]
```
2uasimojo, line 322: Would you mind adding a comment here? I had to go poke around for a while to
- understand why this logic can't be handled in `get_prev_operator_version` itself (though I wonder if it would be worth using some special value, like `- -`, instead of the empty string)
- validate that these empty strings do The Right Thing down the line

2uasimojo (commented):
Thanks for the fix!

Definitely want @bdematte to have a look at this as well.

boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
slaskawi: Ok :)

boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
slaskawi: Comment added.

slaskawi:
All comments addressed. This Pull Request is ready for the next review.

boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
2uasimojo: Thanks.

2uasimojo (approved):
@bdematte and/or @jharrington22 could you please give this the final look?

2uasimojo:
@slaskawi Sorry this hung out so long. I'm going to go ahead and land it. It's harmless at worst, and AFAIK we only have one consumer using OPM anyway.

/lgtm

PR #119, opened 2021-03-15
Integrating pending comment from MR#101

bdematte:


boilerplate/openshift/golang-osd-operator/csv-generate/README.md
```diff
@@ -0,0 +1,46 @@
+# CSV genration and catalog image management
+
+## Scope
+
+The `csv-generate` teagets are covering the historical artifact build for operators to get deployed by OLM : 
+- Generate the CSV (based on python script)
+- Generate the catalog image embeding the newly build CSV
+- Push the generated artifacts to `saas-bundle` gitlab repository (2 branches : `stating` and `prod`)
```
2uasimojo, line 8:
```suggestion
- Push the generated artifacts to `saas-bundle` gitlab repository (2 branches : `staging` and `production`)
```

boilerplate/openshift/golang-osd-operator/csv-generate/README.md
```diff
@@ -0,0 +1,46 @@
...
+VERSION_MINOR=1
+``` 
+
+### CSV compare 
+
+To ease the migration to the common script, a dedicated target has been implemented in order to allow to easily compare the currently generated CSV (expected to be generated with the `hack` scipt) with the common script CSV. 
+It requires to be generated in that order to allow custom-versioning of the CSV (as the version calculation is made internally for the historical `hack` script). 
+
+In case of successful comparison, output should be something similar to : 
+
```
2uasimojo, line 43: nts: WIP here

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -23,12 +23,16 @@ COMMIT_NUMBER=$(shell git rev-list `git rev-list --parents HEAD | egrep "^[a-f0-
...
+VERSION_BASE=$(VERSION_MAJOR).$(VERSION_MINOR)
 OPERATOR_IMAGE=$(IMAGE_REGISTRY)/$(IMAGE_REPOSITORY)/$(IMAGE_NAME)
 OPERATOR_IMAGE_TAG=v$(OPERATOR_VERSION)
 IMG?=$(OPERATOR_IMAGE):$(OPERATOR_IMAGE_TAG)
 OPERATOR_IMAGE_URI=${IMG}
 OPERATOR_IMAGE_URI_LATEST=$(IMAGE_REGISTRY)/$(IMAGE_REPOSITORY)/$(IMAGE_NAME):latest
-OPERATOR_DOCKERFILE ?=build/Dockerfile
+OPERATOR_DOCKERFILE?=build/Dockerfile
+
+# Default value for the CSV generator to be used
```
2uasimojo, line 34: Be nice if this comment denoted the correct override value (`common`).

2uasimojo (commented):
I really only skimmed this since it's still WIP.

boilerplate/_lib/container-make
```diff
@@ -14,7 +14,7 @@ CONTAINER_ENGINE=$(command -v podman || command -v docker)
 
 # First set up a detached container with the repo mounted.
 banner "Starting the container"
-container_id=$($CONTAINER_ENGINE run -d -v "$REPO_ROOT":"$REPO_ROOT" $IMAGE_PULL_PATH tail -f /dev/null)
+container_id=$($CONTAINER_ENGINE run --privileged=true -d -v "$REPO_ROOT":"$REPO_ROOT" $IMAGE_PULL_PATH tail -f /dev/null)
```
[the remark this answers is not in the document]
bdematte: I did it to try having the possibility to run docker in the docker container, but abandonned at least short term.
I reverted it locally when I say it in the overview of the PR, but forgot to push the change.. Pushing it now

test/case/convention/openshift/golang-osd-operator/07-csv-generate-common
```diff
@@ -0,0 +1,128 @@
...
+make staging-saas-bundle-push
+
+# ToDo : Check there is a new commit on staging branch and it contains the previous_version CSV
+
+colorprint $ORANGE "First build and compare on Production"
+make production-common-csv-build
+
+colorprint $ORANGE "First push on Production"
+set_saas_git_config
+make production-saas-bundle-push
```
2uasimojo, line 84: (throughout) Every time I see one of these `*-push` targets, my brain interprets the `push` as meaning that we're actually pushing to quay -- which of course we don't want to do in a test. I had to go read code to figure out that you meant "push to git origin", and you're using "publish" to mean "push to quay". This is further confused by the fact that `catalog-publish.sh` uses `push_catalog` as the variable name for the "publish" action.

What would make it less confusing IMO is:
- Instead of having to add an option to make the script push, instead use an option (via an env var) to _skip_ the push. That way lines like this could say e.g. `make DRY_RUN=true production-saas-bundle-push`. 
- Always use "push" to refer to the quay side, leaving `git push` as implicit, since you're dry-running that part internally by overriding the `origin` remote.

2uasimojo (commented):
The only crucial thing for now is the handling of `LATEST_IMAGE_TAG`. The rest can be deferred.

PR #121, opened 2021-03-16
Parses REPO_ROOT to put repo in gopath

iamkirkbater:
Using bash substring substitution we parse the current path of the
REPO_ROOT and use that to build a CONTAINER_ROOT which then drops your
operator code into the gopath, allowing things like `make generate` to
work out of the box.

### DIFF
--- a/boilerplate/_lib/container-make
+++ b/boilerplate/_lib/container-make
@@ -12,8 +12,15 @@ source ${0%/*}/common.sh
 CONTAINER_ENGINE=$(command -v podman || command -v docker)
 [[ -n "$CONTAINER_ENGINE" ]] || err "Couldn't find a container engine. Are you already in a container?"
 
-CONTAINER_ROOT="/go/src/${REPO_ROOT##*src}"
-[[ -n "$CONTAINER_ROOT" ]] || err "$_lib couldn't parse container root"
+CONTAINER_ROOT="/go/src/${REPO_ROOT##*/src/}"
+ if [[ $CONTAINER_ROOT == "/go/src/${REPO_ROOT}" ]]
+then
+ # If the container root is the same as the repo root without the substring
+ # removal it didn't take because the path might be set up differently
+ # so let's just reset the container root to the repo root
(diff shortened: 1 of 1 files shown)


boilerplate/_lib/container-make
```diff
@@ -12,15 +12,18 @@ source ${0%/*}/common.sh
 CONTAINER_ENGINE=$(command -v podman || command -v docker)
 [[ -n "$CONTAINER_ENGINE" ]] || err "Couldn't find a container engine. Are you already in a container?"
 
+CONTAINER_ROOT="/go/src/${REPO_ROOT##*src}"
```
2uasimojo, line 15: - Can we rely on `REPO_ROOT` having a `src` component? I don't know how boldly we get to assert "your go directory setup is bogus" if not.
- On that note, should we have a fallback (to the old behavior maybe) if that pattern fails?
- To be safe (paranoid), perhaps we should include the slashes
```suggestion
CONTAINER_ROOT="/go/src/${REPO_ROOT##*/src/}"
```
...you know, in case some repository happens to have the substring `src` in its name.

2uasimojo: > * Can we rely on `REPO_ROOT` having a `src` component? I don't know how boldly we get to assert "your go directory setup is bogus" if not.

Update on this: apparently it is not only *not* bogus, it is *recommended* not to put your sources under `GOPATH/src` when using modules (for reasons I don't fully understand). So yeah, we need to come up with something different here.

2uasimojo:
> Hey @iamkirkbater - Good stuff, but for some reason when testing this locally with cloud-ingress-operator and running `./boilerplate/_lib/container-make generate`, it's changing the REPO_ROOT to `cloud-ingress-operator` when it should be `github.com/openshift/cloud-ingress-operator`. Any idea?

What's the abspath to your local clone?

2uasimojo:
Built on this fix via #122. @iamkirkbater if you can confirm that's working for you, we can close this one. Thanks for getting it started.

PR #122, opened 2021-03-17
Fix container-make for openapi-gen

2uasimojo:
`openapi-gen` is really picky about how `GOPATH` is set with respect to the path to the repo clone it's working on. And CI is really picky about where things like `go mod download` are allowed to write. With this change, we make sure the path inside `container-make`'s container gels with `GOPATH` so `openapi-gen` works as expected.

Parts of this were stolen from #121 -- thanks @iamkirkbater.

Co-Authored-By: Kirk Bater <kbater@redhat.com>

### DIFF
--- a/boilerplate/_lib/common.sh
+++ b/boilerplate/_lib/common.sh
@@ -42,11 +42,26 @@ grpcurl_version() {
     $grpcurl -version 2>&1 | cut -d " " -f 2
 }
 
+## repo_import REPODIR
+#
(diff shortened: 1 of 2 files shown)


dustman9000:
I've verified this works locally and there are no diffs with `openapi-generate`... nice work
/lgtm

iamkirkbater:
This is a much cleaner approach than mine :)  Thanks Eric!

PR #129, opened 2021-04-08
Fixing the bad variable used for calling the generate scripts

bdematte:


2uasimojo:
I believe this is correct.

It subsumes all the references to `OPERATOR_IMAGE` in the new code.

The argument to both the common and per-repo generate-operator-bundle.py scripts is what's used to populate the image reference.

The only thing I might suggest is changing the variable name in the common-generate-operator-bundle.py script to `operator_image_uri` -- but meh.

/lgtm
/hold

Unhold at will.

jharrington22:
Validated changes locally.

/lgtm

PR #133, opened 2021-04-09
Enable container-make in boilerplate itself

2uasimojo:
Previously in order for `container-make` to function in the boilerplate repository itself, you had to set the `LATEST_IMAGE_TAG` environment variable manually. With this commit, if it's unset and we're in boilerplate, we discover it from git.


2uasimojo:
This has been tested via https://github.com/openshift/deadmanssnitch-operator/pull/107 and works like a charm. Ready to go.

PR #136, opened 2021-04-12
Generate operator bundle generically and inclusively

2uasimojo:
This is an almost-total rewrite of the common operator bundle generator script to accommodate variations in boilerplate consumers.

Before: The script looked for specific files by name. This wasn't a big deal as long as consumers had the right corresponding file they could simply rename. However, it would blow up if those files weren't present or didn't contain exactly what was expected; or (importantly) if more than one file of that kind was needed.

After: The script processes all yaml documents under the deploy/ subdirectory, regardless of filename, including cases where multiple yaml documents exist in a single file, and processes them based on their Kind. The following are treated specially:

- CustomResourceDefinitions: These are registered as "owned" in the CSV. Otherwise they are copied into the bundle directory as usual.
- ClusterRoles: These are folded into the CSV's `clusterPermissions` and *not* copied into the bundle directory.
- ClusterRoleBindings: These are only used to discover the appropriate `serviceAccountName` for their corresponding ClusterRoles. They are not copied into the bundle directory. If any ClusterRoleBindings are "left over" at the end (not used to find a ServiceAccount), an error is raised.
- Deployment: We explicitly validate that there is exactly one of these. We fold the contents into the CSV's `deployments`. The Deployment is *not* copied to the bundle directory.
- ServiceAccounts: These are *ignored*, as we assume they correspond exactly to those listed in ClusterRoleBindings or the Deployment's Pod spec.

TODO:s
- The CSV [schema](https://docs.openshift.com/container-platform/4.5/rest_api/operatorhub_apis/clusterserviceversion-operators-coreos-com-v1alpha1.html) appears to support Role[Binding]s in a manner similar to ClusterRole[Binding]s. Assuming OLM supports that, we should use it. For now, this commit treats Role[Binding]s as generic resources and simply copies them into the bundle, as consumers appeared to be doing.
- We should validate ServiceAccounts to make sure they correspond to [Cluster]RoleBindings or Pod specs in the Deployment, and error if any are "left over".

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
+++ b/boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
@@ -22,13 +22,13 @@ VERSION_BASE = "0.1"
 parser = argparse.ArgumentParser()
 parser.add_argument("-o", "--operator-name", type=str, help="Name of the operator", required=True)
 parser.add_argument("-d", "--output-dir", type=str, help="Directory for the CSV generation", required=True)
-parser.add_argument("-p", "--previous-version", type=str, help="Directory for the CSV generation", required=True)
+parser.add_argument("-p", "--previous-version", type=str, help="Semver of the version being replaced", required=True)
 parser.add_argument("-n", "--commit-number", type=str, help="Number of commits in the project (used for version generation)", required=True)
 parser.add_argument("-c", "--commit-hash", type=str, help="Current commit hashDirectory for the CSV generation (used for version generation)", required=True)
 parser.add_argument("-i", "--operator-image", type=str, help="Base index image to be used", required=True)
 args = parser.parse_args()
 
-operator_name   = args.operator_name
+OPERATOR_NAME   = args.operator_name
 outdir          = args.output_dir
 prev_version    = args.previous_version
 git_num_commits = args.commit_number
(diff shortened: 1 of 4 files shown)


boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
```diff
@@ -41,25 +41,159 @@
...
-        shutil.copy(full_path, os.path.join(version_dir, file_name))
-    # Load CRD so we can use attributes from it
-    with open("deploy/crds/{}".format(file_name), "r") as stream:
-        crd = yaml.load(stream)
-    # Update CSV template customresourcedefinitions key
+VERSION_DIR = os.path.join(outdir, full_version)
+if not os.path.exists(VERSION_DIR):
+    os.mkdir(VERSION_DIR)
+
+with open('config/templates/csv-template.yaml'.format(OPERATOR_NAME), 'r') as stream:
```
arjunrn, line 48: The `format()` call here does nothing.

boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
```diff
@@ -69,43 +203,81 @@
...
         }
     )
+# These will be written to the bundle at the end along with generic resources
 
-csv['spec']['install']['spec']['clusterPermissions'] = []
+## Process [Cluster]Role[Binding]s (TODO: Match up ServiceAccounts)
+# Role and ClusterRole are processed similarly
+rolekind_csvkey_bindingmap = (
+    ('ClusterRole', 'clusterPermissions', crb_by_cr),
+    # NOTE: The schema supports `permissions` for Roles, but OLM may not. For
```
arjunrn, line 212: > but OLM may not

Has this been confirmed?

boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
```diff
@@ -69,43 +203,81 @@
...
+        # bundle resources.
+        trim_index(by_kind, f"{kind}Binding", binding)
+        # Inject the rules and ServiceAccount
+        csv['spec']['install']['spec'][csvkey].append(
+            {
+                'rules': role['rules'],
+                'serviceAccountName': sa,
+            })
+    # Get rid of these so we can iterate over what's left at the end
+    trim_index(by_kind, kind, 'ALL')
```
arjunrn, line 235: I'm not a 100% sure but OLM may support provisioning `ClusterRoles` for service accounts in other namespaces not associated with the operator. I'm certain it does so for `Roles`. So I would recommend only trimming CRs and CRBs which are associated with the operator service account.

boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
2uasimojo: True story. I commented on that elsewhere as well. Fixed.

boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
2uasimojo: No. The point of this comment is that we need to find out. For now we're doing things the same way as the operators have been. IMO we shouldn't try to change that in this PR.

boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
2uasimojo: We're already integrating all the CRs and CRBs by folding them into the CSV. The schema does not have a field for ServiceAccount namespace, so I don't see how that could be supported. But you raise a good point about the Roles and RoleBindings: At the moment we're just putting those in the bundle but ignoring their ServiceAccounts. I imagine we need to put those in the bundle as well.

I wonder if redundant ServiceAccounts (i.e. listed in a `clusterPermissions` entry and also in a bundle file) hurt anything. Anyway, I can go ahead with the work to pare down all SAs that we've already "used" and just bundle the rest.

2uasimojo: (Later) On further thought & discussion, I don't believe we're ever creating ServiceAccounts in other namespaces -- so let's leave this alone unless you find a counterexample.

boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
2uasimojo: Upon further thought & discussion:
Whether or not OLM actually supports the `permissions` allowed by the schema, it doesn't have a way to manage cross-namespace bindings -- and that's why we're seeing these files being laid down explicitly in the bundle.

We should possibly queue up a TODO to introspect RoleBindings we come across, determine whether they reference owned ServiceAccounts, and include them in `permissions` if they do. But for now, unless we come across a case of this in an existing operator, let's leave it as written (*all* Role[Binding]s go into the bundle directly).

boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
```diff
@@ -41,25 +41,159 @@
...
+
+def filename_for(resource):
+    """Generate a unique base file name for the given resource.
+
+    :param resource: A k8s resource definition dict, expected to have at least
+        a kind and metadata.name.
+    :return: A base file name (no directory) that should be unique for the
+        given resource (assuming the resource is itself unique).
+    """
+    return '_'.join((resource['kind'], resource['metadata']['name'])) + '.yaml'
```
jharrington22, line 116: Slight concern that this resource may overwrite another if its the same kind+name but different namespace. Include namespace if available or append unique identifier?

jharrington22: Or possibly just check if the path already exists in `bundle()` and error.

boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
```diff
@@ -41,25 +41,159 @@
...
+        index.pop(kind, None)
+        return
+    # Find the specific resource to remove
+    for i in range(len(index[kind])):
+        if index[kind][i]['metadata']['name'] == item['metadata']['name']:
+            found = i
+            break
+    else:
+        # not found
+        return
```
jharrington22, line 178: We expect there to only every be a single item with `.metadata.name` of matching `kind` I _think_ this is ok but it would fail silently.

boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
```diff
@@ -41,25 +41,159 @@
...
+if 'Deployment' not in by_kind:
+    raise NoDeploymentFound()
+
+# TODO: Should we support additional Deployments that aren't the operator
+# Deployment?
+if len(by_kind['Deployment']) > 1:
+    raise MultipleDeploymentsNotSupported(by_kind['Deployment'])
+
+## Process CRDs
+if 'CustomResourceDefinition' in by_kind:
```
jharrington22, line 192: This is so much nicer.

boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
jharrington22: I confirmed that none of our CSV templates reference `permissions` 

```
find ./* -type f -regextype egrep -regex ".*csv-template.*" -exec grep -iH permissions: {} \;
./aws-account-operator/config/templates/aws-account-operator-csv-template.yaml:      clusterPermissions:
./aws-efs-operator/config/templates/aws-efs-operator-csv-template.yaml:      clusterPermissions:
./certman-operator/config/templates/certman-operator-csv-template.yaml:      clusterPermissions:
./custom-domains-operator/config/templates/custom-domains-operator-csv-template.yaml:      clusterPermissions:
./deadmanssnitch-operator/config/templates/deadmanssnitch-operator-csv-template.yaml:      clusterPermissions:
./gcp-project-operator/config/templates/gcp-project-operator-csv-template.yaml:      clusterPermissions:
./managed-upgrade-operator/config/templates/managed-upgrade-operator-csv-template.yaml:      clusterPermissions:
./must-gather-operator/config/templates/must-gather-operator-csv-template.yaml:      clusterPermissions:
./osd-metrics-exporter/config/templates/osd-metrics-exporter-csv-template.yaml:      clusterPermissions:
./rbac-permissions-operator/config/templates/rbac-permissions-operator-csv-template.yaml:      clusterPermissions:
./splunk-forwarder-operator/config/templates/splunk-forwarder-operator-csv-template.yaml:      clusterPermissions:
```

Managed velero seems to have two CSV templates 

```
./managed-velero-operator/config/templates/managed-velero-operator-csv-template.yaml:      clusterPermissions:
./managed-velero-operator/config/templates/managed-velero-operator-csv-template.yaml:      permissions:
```

And configure-alertmanager-operator needs the path of the template changed and the permissions are added in the generate script [here](https://github.com/openshift/configure-alertmanager-operator/blob/master/build/generate-operator-bundle.py#L95)

```
cat build/templates/csv.yaml.tmpl
apiVersion: operators.coreos.com/v1alpha1
kind: ClusterServiceVersion
```

boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
jharrington22: From what I can see on cluster and other operators OLM doesn't create ServiceAccounts in namespaces other than the operators install namespace.

boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
2uasimojo: Yes, good call. I actually started doing that this morning when I was working on the ServiceAccount paring thing we decided not to do. I'll resurrect it.

boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
2uasimojo: I think we're concerned that it would erroneously remove the wrong item. I'll add match by ns on L173.

PR #139, opened 2021-04-14
Ensure we fetch a valid deployed production hash

jharrington22:
This PR does two things to discover and ensure we fetch a valid HASH.

1. We determine if the SAAS file is applying a resource type of `SelectorSyncSet` if it is we'll assume the saas file is for a customer cluster and not Hive. (Resources cannot be sync'd to a customer cluster without a SelectorSyncSet (I should validate that and SyncSet).
2. Once we know the resource type we can make an assumption on the namespace path reference in resourceTemplates.

Before adding validation (Non hive cluster):

```
Image quay.io/jharrington22/cloud-ingress-operator-registry:production-1a3ab03 was deleted from the repository.
Proceeding as if it never existed.
make[1]: Entering directory '/home/jamesh/.gvm/pkgsets/go1.13.6/global/src/github.com/openshift/cloud-ingress-operator'
Cloning into 'saas-cloud-ingress-operator-bundle'...
remote: Enumerating objects: 155, done.
remote: Counting objects: 100% (155/155), done.
remote: Compressing objects: 100% (50/50), done.
remote: Total 1348 (delta 97), reused 138 (delta 81), pack-reused 1193
Receiving objects: 100% (1348/1348), 130.62 KiB | 228.00 KiB/s, done.
Resolving deltas: 100% (891/891), done.
SAAS file is NOT applied to Hive, MANAGED_RESOURCE_TYPE=SelectorSyncSet
Current deployed production HASH: 
Error discovering current production deployed HASH
make[1]: *** [boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.mk:42: production-common-csv-build] Error 1
make[1]: Leaving directory '/home/jamesh/.gvm/pkgsets/go1.13.6/global/src/github.com/openshift/cloud-ingress-operator'
make: *** [boilerplate/openshift/golang-osd-operator/standard.mk:211: build-push] Error 2

```

After (Non hive cluster)

```
Image quay.io/jharrington22/cloud-ingress-operator-registry:staging-1a3ab03 exists with digest sha256:64b5c9d1090c21f67419a282ea99971d62ab23ea3e53c05e2c4bae622d09e0ec.                                                                                       
Catalog image quay.io/jharrington22/cloud-ingress-operator-registry:staging-1a3ab03 already                                                                                                                                                                   
exists. Assuming this means the saas bundle work has also been done                                                                                                                                                                                           
properly. Nothing to do!                                                                                                                                                                                                                                      
Image quay.io/jharrington22/cloud-ingress-operator-registry:production-1a3ab03 was deleted from the repository.                                                                                                                                               
Proceeding as if it never existed.                                                                                                                                                                                                                            
make[1]: Entering directory '/home/jamesh/.gvm/pkgsets/go1.13.6/global/src/github.com/openshift/cloud-ingress-operator'                                                                                                                                       
Cloning into 'saas-cloud-ingress-operator-bundle'...                                                                                                                                                                                                          
remote: Enumerating objects: 155, done.                                                                                                                                                                                                                       
remote: Counting objects: 100% (155/155), done.                                                                                                                                                                                                               
remote: Compressing objects: 100% (50/50), done.                                                                                                                                                                                                              
remote: Total 1348 (delta 97), reused 138 (delta 81), pack-reused 1193                                                                                                                                                                                        
Receiving objects: 100% (1348/1348), 130.62 KiB | 165.00 KiB/s, done.                                                                                                                                                                                         
Resolving deltas: 100% (891/891), done.                                                                                                                                                                                                                       
SAAS file is NOT applied to Hive, MANAGED_RESOURCE_TYPE=SelectorSyncSet                                                                                                                                                                                       
Current deployed production HASH: 10af8dab0515ed5762aa283c83eb4080b0238fcd                                                                                                                                                                                    
Generating CSV for version: 0.1.344-1a3ab03    
```

We can go further to validate this by running the built catalog images and comparing the versions returned by the registry API and what we expect according to the `saas-${operator_name}-bundle` repo. This is how I realized the issue in the first place.

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
+++ b/boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
@@ -4,7 +4,7 @@ source `dirname $0`/common.sh
 
 usage() { echo "Usage: $0 -o operator-name -c saas-repository-channel -H operator-commit-hash -n operator-commit-number -i operator-image -V operator-version -g [hack|common][-d]" 1>&2; exit 1; }
 
-# TODO : Add support of long-options 
+# TODO : Add support of long-options
 while getopts "c:dg:H:i:n:o:V:" option; do
     case "${option}" in
         c)
@@ -51,7 +51,7 @@ check_mandatory_params operator_channel operator_image operator_version operator
 
 # Use set container engine or select one from available binaries
 if [[ -z "$CONTAINER_ENGINE" ]]; then
-	CONTAINER_ENGINE=$(command -v podman || command -v docker || true)
+    CONTAINER_ENGINE=$(command -v podman || command -v docker || true)
 fi
 
 if [[ -z "$CONTAINER_ENGINE" ]]; then
@@ -69,7 +69,7 @@ fi
 REPO_DIGEST=${operator_image}@${IMAGE_DIGEST}
 
 # If no override, using the gitlab repo
-if [ -z "$GIT_PATH" ] ; then 
+if [ -z "$GIT_PATH" ] ; then
     GIT_PATH="https://app:@gitlab.cee.redhat.com/service/saas-${operator_name}-bundle.git"
 fi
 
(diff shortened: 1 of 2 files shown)


2uasimojo (commented):
Doesn’t it depend whether we’re building a hive operator vs a cluster operation which way this needs to go?

maorfr:
https://github.com/openshift/aws-account-operator/pull/493

2uasimojo:
I guess the concern with switching to graphql is that it requires yet another auth token for local testing. And iirc that one isn’t trivial to get — requires a MR in app-interface and pgp and...

maorfr:
And that's it.

jharrington22:
> I guess the concern with switching to graphql is that it requires yet another auth token for local testing. And iirc that one isn’t trivial to get — requires a MR in app-interface and pgp and...

Exactly. This works as is and gives us _some_ safety. ++ to switching to graphql but WHEN we have boilerplate we can do it all at once.

maorfr:
App-interface being in gitlab is an implementation detail you shouldn't rely on. It also restricts us from moving it, which we may consider in the future. It also forces us to run these jobs behind the VPN so the clone will work.

Worse, in case we decide to change the schema structure, we suddenly have a fire on our hands as all your jobs will start to fail.
The (graphql) API is what you SHOULD be using.

The concern about the MR - it takes 2 minutes. And you should probably have your gpg keys in app-interface as part of your onboarding process. It has many other usages that you may need to use in the future.

I can go on forever 😉

jharrington22:
> App-interface being in gitlab is an implementation detail you shouldn't rely on. It also restricts us from moving it, which we may consider in the future. It also forces us to run these jobs behind the VPN so the clone will work.
> 
> Worse, in case we decide to change the schema structure, we suddenly have a fire on our hands as all your jobs will start to fail.
> The (graphql) API is what you SHOULD be using.
> 
> The concern about the MR - it takes 2 minutes. And you should probably have your gpg keys in app-interface as part of your onboarding process. It has many other usages that you may need to use in the future.
> 
> I can go on forever

1000% agree @maorfr are the ENV credentials that we need already exist? Mostly I'm trying to change as few pieces as possible while we migrate to boilerplate everywhere. I'm pretty sure that a schema change is the cause of all our OLM bugs in prod (will link later) so I want to get here as soon as we can.

maorfr:
Everything is ready and a working example is the PR I mentioned above.

maorfr:
For the sake of splitting efforts, this change is LGTM, and a switch to graphql is not urgent at all.

jharrington22:
@maorfr if everything is there I'm all in :slightly_smiling_face: I thought we had to make a couple of MR's get what we need available to the job.

2uasimojo:
> @maorfr if everything is there I'm all in  I thought we had to make a couple of MR's get what we need available to the job.

The *job* already has what it needs. It's local testing that would be painful. So I'm good with it as long as the PR that cuts over to graphql also updates the [testing doc](https://github.com/openshift/boilerplate/blob/f2167d69d378ebb468a49e54f2cae068ab10a383/boilerplate/openshift/golang-osd-operator/app-sre.md) to describe how to set up those creds. Most of that is covered in [these instructions](https://gitlab.cee.redhat.com/service/app-interface#querying-the-app-interface), but it would be useful to also a) point to an [example MR](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/10873) and b) go into a little more detail about how to configure the creds in the environment (I recall struggling with that a little bit).

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
```diff
@@ -85,9 +85,21 @@ else
         if [ -z "$DEPLOYED_HASH" ] ; then
             DEPLOYED_HASH=$(
                 curl -s "https://gitlab.cee.redhat.com/service/app-interface/raw/master/data/services/osd-operators/cicd/saas/saas-${operator_name}.yaml" | \
-                    docker run --rm -i quay.io/app-sre/yq:3.4.1 yq r - "resourceTemplates[*].targets(namespace.\$ref==/services/osd-operators/namespaces/hivep01ue1/${operator_name}.yml).ref"
+                    docker run --rm -i quay.io/app-sre/yq:3.4.1 yq r - "resourceTemplates[*].targets(namespace.\$ref==/services/osd-operators/namespaces/hivep01ue1/cluster-scope.yml).ref"
```
2uasimojo, line 88: Did we ascertain that this will work for all the operators? Based on what's in [that directory](https://gitlab.cee.redhat.com/service/app-interface/-/tree/master/data/services/osd-operators/namespaces/hivep01ue1), I'm guessing not -- we would need to condition cluster- vs hive-scoped operators somehow.

jharrington22:
@2uasimojo lets merge as is, I'll create a card to follow up and change it to graphql once we have all the pieces sorted.

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
jharrington22: let me validate.

2uasimojo:
> @2uasimojo lets merge as is, I'll create a card to follow up and change it to graphql once we have all the pieces sorted.

I think we crossed in the mail -- please see above [comment](https://github.com/openshift/boilerplate/pull/139#discussion_r613386684)

You would also need to fix CI :)

jharrington22:
Yeah @2uasimojo I need to update this concludes our deploy issues, its not the same. 

/hold

jharrington22:
Testing this solution

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
```diff
@@ -79,14 +79,66 @@ if [ "$diff_generate" = true ] ; then
...
+    if [[ "${MANAGED_RESOURCE_TYPE}" == "" ]]; then
+        echo "Unabled to determine if SAAS file managed resource type"
+        exit 1
+    fi
+
+    # Output our result and make it easier to read by defining a Hive bool
+    HIVE=false
+    if [[ "${MANAGED_RESOURCE_TYPE}" == "SelectorSyncSet" ]]; then
+        echo "SAAS file is NOT applied to Hive, MANAGED_RESOURCE_TYPE=$MANAGED_RESOURCE_TYPE"
+        HIVE=true
```
2uasimojo, line 109:
```suggestion
        HIVE=false
```
?

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
```diff
@@ -79,14 +79,66 @@ if [ "$diff_generate" = true ] ; then
...
+    else
+        echo "SAAS file is applied to Hive, MANAGED_RESOURCE_TYPE=$MANAGED_RESOURCE_TYPE"
+        HIVE=true
+    fi
     
     # remove any versions more recent than deployed hash
+    # TODO: Combine the above and below if statements in a single YQ statement
+    # I couldn't seem to make it work with yq 3.4.1
```
2uasimojo, line 117: Why not just set a variable?
```bash
if $HIVE; then
    filename=${operator_name}.yml
else
    filename=cluster-scope.yml
fi
...
docker run ... "...\$ref==/services/osd-operators/namespaces/hivep01ue1/${filename}).ref"
```

2uasimojo (changes requested):
This looks workable except for the mistake noted inline

but

Perhaps it would be easier to require the consumer to set a Makefile variable indicating whether they're hive or not. The less we can rely on specific yaml formats (or even graphql queries) the better.

2uasimojo (approved):

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
jharrington22: Good idea.

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
jharrington22: Fixed.

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
jharrington22: redundant now.

jharrington22: > This looks workable except for the mistake noted inline
> 
> but
> 
> Perhaps it would be easier to require the consumer to set a Makefile variable indicating whether they're hive or not. The less we can rely on specific yaml formats (or even graphql queries) the better.

I don't disagree with this. We have protection in that if we guess wrong the hash should be blank and we'll error. In the interest of time I think lets move forward with this (since its only one merge and we don't have to touch the other operators) and fix this later. https://issues.redhat.com/browse/OSD-6995

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
```diff
@@ -79,15 +79,58 @@ if [ "$diff_generate" = true ] ; then
...
+    # are being applied to hive.
+    # For customer cluster resources this should always be of type "SelectorSyncSet" resources otherwise
+    # can't be sync'd to the customer cluster. We're explicity selecting the first element in the array.
+    # We can safely assume anything that is not of type "SelectorSyncSet" is being applied to hive only
+    # since it matches ClusterDeployment resources.
+    # From this we'll assume that the namespace reference in resourceTemplates to be:
+    # For customer clusters: /services/osd-operators/namespace/<hive shard>/namespaces/cluster-scope.yaml
+    # For hive clusters: /services/osd-operators/namespace/<hive shard>/namespaces/<namespace name>.yaml
+    MANAGED_RESOURCE_TYPE=$(curl -s "${SAAS_FILE_URL}" | \
+            docker run --rm -i quay.io/app-sre/yq:3.4.1 yq r - "managedResourceTypes[0]"
```
2uasimojo, line 98: `docker` will need to be conditioned. Something like [this](https://github.com/openshift/boilerplate/pull/142/files#diff-424ddf34b68cef0633f4520f756a919d70a24880b5ab931d3cd85d17f8659b5fR119-R126) :P

jharrington22: Fixed.

2uasimojo:
This should be rebased now that #142 has merged.

The assignment of `CONTAINER_ENGINE` should move up to where you've got it.

And every invocation that might be containerized will need to be conditioned on that variable being set or not.

2uasimojo (approved):
Nice

/lgtm

2uasimojo:
Okay, so CI was failing because the gitlab query was coming up empty. We thought it was because we were using a dummy project that didn't have a SaaS file in app-interface. So I first tried adding an override to point the `SAAS_FILE_URL` to a known project's SaaS file.

But it still didn't work -- the curl command was producing no output and exiting with RC=60 ("Peer certificate cannot be authenticated with known CA certificates.")

Turns out the CI environment doesn't have the right certificates to curl gitlab (or maybe anything?) over https. So I *also* added a granny switch to curl with `--insecure` and enabled it in the test case.

2uasimojo:
I'm going to stamp this so it lands if CI passes.

/lgtm

This will unblock us and allow us to propose bp updates in consumer projects.

2uasimojo:
Okay, at this point I'm guessing it's just that the CI env doesn't have access to gitlab at all (like, it's not inside the VPN?). So I'm going back to disabling that whole chunk.

2uasimojo:
And

/lgtm

again, for the [same reasons](https://github.com/openshift/boilerplate/pull/139#issuecomment-820074613).

PR #140, opened 2021-04-14
Set the operator bundle generator python script to use python

arjunrn:
`python3` may not be defined in some environments

[efried] To be clear, the script still relies on the python version being 3; we’re just no longer being explicit about it because apparently build environments are defaulting “python” to 3 *and* starting to not provide the “python3” alias/symlink.

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
+++ b/boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 #
 # Common script to generate OSD operator bundles for publishing to OLM. Copies appropriate files
 # into a directory, and composes the ClusterServiceVersion which needs bits and


boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
```
2uasimojo, line 1: I’ve never loved the “env” way of shebanging. Does the Jenkins server have /usr/bin/python3? I sure would love to remain explicit if that’s possible.

arjunrn: You can also change how the script is started to `python3 gen...` instead. That is more likely to work than the `env` approach.

2uasimojo: Even better, let's run the script in a py3 container.

(We've just learned py3 doesn't even exist on the jenkins server :mindblown:)

2uasimojo (changes requested):
This is [failing CI due to py3-specific syntax](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_boilerplate/140/pull-ci-openshift-boilerplate-master-pr-check/1382310517950910464#1:build-log.txt%3A3999) because our backing image defaults to py2. Per [conversation with appsre](https://coreos.slack.com/archives/CCRND57FW/p1618406606105400?thread_ts=1618404527.098900&cid=CCRND57FW), the same is true of (at least some of) the jenkins environment(s).

So let's find a py3 container (where? there's no quay.io/app-sre/python*) in which to invoke the script explicitly.

/hold

arjunrn:
Closing in favor of #142

PR #143, opened 2021-04-14
Small improvements to make it easier to run csv-generate test locally

arjunrn:
- When the dummy `skopeo` is not the first in the path the actual `skopeo` executable will be used causing the test to fail.
- When a git repo is initialized the name of the default branch is taken from the users local configuration which could be something other than `master`. So set it explicitly.

### DIFF
--- a/test/case/convention/openshift/golang-osd-operator/06-csv-generate
+++ b/test/case/convention/openshift/golang-osd-operator/06-csv-generate
@@ -29,7 +29,7 @@ cat <<EOF >/tmp/bin/skopeo
 echo '{"Digest": "sha256:abc123"}'
 EOF
 chmod +x /tmp/bin/skopeo
-export PATH=$PATH:/tmp/bin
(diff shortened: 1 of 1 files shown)


test/case/convention/openshift/golang-osd-operator/06-csv-generate
```diff
@@ -29,7 +29,7 @@ cat <<EOF >/tmp/bin/skopeo
 echo '{"Digest": "sha256:abc123"}'
 EOF
 chmod +x /tmp/bin/skopeo
-export PATH=$PATH:/tmp/bin
+export PATH=/tmp/bin:$PATH
```
2uasimojo, line 32: Yeah, good call, I should have thought of that.

2uasimojo (approved):
/lgtm

jharrington22 (approved):
/lgtm

PR #149, opened 2021-04-16
Error when unsupported resources present in deploy directory

arjunrn:
Only select resources are supported in the OLM package manifest format. Other types should cause an exception.

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
+++ b/boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
@@ -19,6 +19,28 @@ import shutil
 import argparse
 import string
 
+# The registry is pinned to version 4.7 and only the following resouces are permitted in
+# the bundle. The full list can be found at https://github.com/operator-framework/operator-registry/blob/release-4.7/pkg/lib/bundle/supported_resources.go#L4-L19
+BUNDLE_PERMITTED_RESOURCES = (
+    "ClusterServiceVersion",
+    "CustomResourceDefinition",
+    "Deployment", # this resource is injected into the CSV directly but not supported by the registry
+    "Secret",
+    "ClusterRole",
+    "ClusterRoleBinding",
+    "ConfigMap",
+    "ServiceAccount",
+    "Service",
+    "Role",
+    "RoleBinding",
+    "PrometheusRule",
+    "ServiceMonitor",
+    "PodDisruptionBudget",
+    "PriorityClass",
+    "VerticalPodAutoscaler",
+    "ConsoleYamlSample",
+)
+
 parser = argparse.ArgumentParser()
 parser.add_argument("-o", "--operator-name", type=str, help="Name of the operator", required=True)
 parser.add_argument("-d", "--output-dir", type=str, help="Directory for the CSV generation", required=True)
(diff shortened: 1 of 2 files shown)


boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
```diff
@@ -19,6 +19,25 @@
 import argparse
 import string
 
+BUNDLE_PERMITTED_RESOURCES = (
+    "ClusterServiceVersion"
```
jharrington22, line 23: Erroneous space.

boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
```diff
@@ -19,6 +19,25 @@
 import argparse
 import string
 
+BUNDLE_PERMITTED_RESOURCES = (
```
2uasimojo, line 22: Can we include a comment linking to the source for this?

https://github.com/operator-framework/operator-registry/blob/release-4.7/pkg/lib/bundle/supported_resources.go#L3-L20

boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
```diff
@@ -310,6 +329,9 @@ def trim_index(index, kind, item):
 ## Process any additional resources
 # These don't go in the CSV; they just get copied into the bundle as is.
 for kind, docs in by_kind.items():
+    if kind not in BUNDLE_PERMITTED_RESOURCES:
+        print(f'Skipping resource of kind {kind}')
+        continue
```
2uasimojo, line 334: Let's error here to force the user to have all and only supported resources in their deploy folder. Otherwise they might expect them to do something. Especially since we've made it easy to test the pipeline locally, this shouldn't be an undue burden.

(And if we're going to do that, we could put this loop up top for efficiency)

2uasimojo (commented):
This is awesomeness. Thanks @arjunrn

boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
2uasimojo: Also useful in the comment would be to note that we [pinned the registry version to 4.7.0](https://github.com/openshift/boilerplate/blob/master/boilerplate/openshift/golang-osd-operator/csv-generate/catalog-build.sh#L57), so upgrading to support new kinds would entail bumping that version as well as this list.

boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
2uasimojo: Don't you need commas in here? Is this even compiling?

boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
```diff
@@ -65,6 +86,8 @@
                 print(
                     f"Loading {doc['kind']} {doc['metadata']['name']} " +
                     f"from {path}")
+                if doc['kind'] not in BUNDLE_PERMITTED_RESOURCES:
+                    raise UnsupportedRegistryResourceKind(doc['kind'], path)
```
2uasimojo, line 90: I'm testing this locally rn, but I think this will blow up because `UnsupportedRegistryResourceKind` isn't defined yet.

/hold

2uasimojo: Okay, yeah, couple problems here.
1. As noted, this [raises `NameError` rather than `UnsupportedRegistryResourceKind`](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_boilerplate/149/pull-ci-openshift-boilerplate-master-pr-check/1383065847995568128#1:build-log.txt%3A4008) because the latter isn't defined yet.
2. It's [raising it on the `Deployment` resource](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_boilerplate/149/pull-ci-openshift-boilerplate-master-pr-check/1383065847995568128#1:build-log.txt%3A4004). We should add that to the supported list with a comment that it's processed specially by the script and not included in the bundle.

2uasimojo:
Thanks Arjun.

/lgtm

PR #152, opened 2021-04-16
removes debug flag from container-make

iamkirkbater:
Removes the debug flag on container-make.  If still needed by end-users, they can always set it in their session.

[efried] ...via `BOILERPLATE_SET_X=1`

### DIFF
--- a/boilerplate/_lib/container-make
+++ b/boilerplate/_lib/container-make
@@ -1,4 +1,4 @@
-#!/bin/bash -x
+#!/bin/bash
 
 if [[ "$1" == "-h"* ]] || [[ "$1" == "--h"* ]]; then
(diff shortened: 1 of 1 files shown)


boilerplate/_lib/container-make
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -x
+#!/bin/bash -e
```
2uasimojo, line 1: We can't add -e as is. The script isn't set up for it. For example, L22 will bail without giving us a chance to check the RC and print a helpful message. Likewise L30 would bail and not let us start the container for you to debug in.

2uasimojo (changes requested):

2uasimojo:
/lgtm
/approve
Thanks Kirk!

PR #153, opened 2021-04-16
Persist codecov script used as a test artifact

cblecker:
This will persist the codecov script used as an artifact of the coverage job

ARTIFACT_DIR is defined here: https://docs.ci.openshift.org/docs/architecture/step-registry/#available-environment-variables

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/codecov.sh
+++ b/boilerplate/openshift/golang-osd-operator/codecov.sh
@@ -33,7 +33,7 @@ elif [[ "${JOB_TYPE}" == "local" ]]; then
        echo "coverage report available at ${COVER_PROFILE}"
        exit 0
 else
-       echo "${JOB_TYPE} jobs not supported"
+       echo "${JOB_TYPE} jobs not supported" >&2
        exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


cblecker:
Tested using the following stub script so I didn't need to go through all the CI stuff with the tokens:
```
#!/usr/bin/env bash

set -o errexit
set -o nounset
set -o pipefail

if [[ -n "${ARTIFACT_DIR:-}" ]]; then
       curl -s https://codecov.io/bash -o "${ARTIFACT_DIR}/codecov.sh"
       bash <(cat "${ARTIFACT_DIR}/codecov.sh") -h
else
       bash <(curl -s https://codecov.io/bash) -h
fi
```

jaybeeunix:
/lgtm

:heart_eyes:

boilerplate/openshift/golang-osd-operator/codecov.sh
```diff
@@ -42,4 +42,9 @@ export CI_BUILD_URL="${JOB_LINK}"
 export CI_BUILD_ID="${JOB_NAME}"
 export CI_JOB_ID="${BUILD_ID}"
```
2uasimojo, line 44: Might we want to check `-d` as well?

boilerplate/openshift/golang-osd-operator/codecov.sh
```diff
@@ -42,4 +42,9 @@ export CI_BUILD_URL="${JOB_LINK}"
...
 
-bash <(curl -s https://codecov.io/bash) -Z -K -f "${COVER_PROFILE}" -r "${REPO_OWNER}/${REPO_NAME}" ${REF_FLAGS}
+if [[ -n "${ARTIFACT_DIR:-}" ]]; then
+       curl -s https://codecov.io/bash -o "${ARTIFACT_DIR}/codecov.sh"
+       bash <(cat "${ARTIFACT_DIR}/codecov.sh") -Z -K -f "${COVER_PROFILE}" -r "${REPO_OWNER}/${REPO_NAME}" ${REF_FLAGS}
+else
```
2uasimojo, line 48: Much as I dig the idea of a failsafe, this opens the possibility of us never storing the script and not finding out about it until we need it. I think it might be better just to blow up, forcing the user to figure out what’s wrong and fix it.

cblecker: This was meant to handle local job types. I've now added additional checks.

2uasimojo (approved):
Dig

/lgtm

PR #154, opened 2021-04-18
Make sed prompt to use compatible sed

cblecker:
BSD and GNU sed have different syntaxes. The BSD sed ships by default on mac.

Needed to duplicate this in a couple places as `boilerplate/update` needs to be standalone, as well as the test harness.


2uasimojo:
Tested this on my (Linux) machine throug a bootstrap and an update. The fact that `pr-check` is passing means that bit also works on Linux. I'm having to assume this makes Macs happy.

/lgtm
/approve

PR #158, opened 2021-04-19
Use /usr/bin/env bash everywhere

cblecker:


config/tag-check.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -xe
+#!/usr/bin/env bash -xe
```
cblecker, line 1: What environment are we running in that doesn't have /usr/bin/env?

cblecker:
@2uasimojo I think this should be ready, but it needs an image pushed it looks like?

2uasimojo:
> @2uasimojo I think this should be ready, but it needs an image pushed it looks like?

It's because you [touched files in the config/ directory](https://github.com/openshift/boilerplate/tree/766ff62ff278b097424d621d9a40a208f41120f5/config). This makes our CI ([pr-check](https://github.com/openshift/boilerplate/blob/eb577a390b8cd99e45be60d635d7ca9cdcab62f3/Makefile#L22) => [test](https://github.com/openshift/boilerplate/blob/eb577a390b8cd99e45be60d635d7ca9cdcab62f3/Makefile#L18) => [tag-check](https://github.com/openshift/boilerplate/blob/eb577a390b8cd99e45be60d635d7ca9cdcab62f3/Makefile#L15) => [tag-check.sh](https://github.com/openshift/boilerplate/blob/eb577a390b8cd99e45be60d635d7ca9cdcab62f3/config/tag-check.sh)) assume you've altered the backing image we build, and if you do that, you're required to produce a new tag, per [these instructions](https://github.com/openshift/boilerplate/blob/master/README.md#build-images).

Since you're not actually changing the backing image, and those scripts almost never need to be run locally, you may just want to revert that part of the PR.

But if you did want to make a new image version, you could figure out how to install skopeo for us...

cblecker:
https://github.com/openshift/release/pull/17916

2uasimojo:
To make CI happy, you'll need to update [this](https://github.com/openshift/boilerplate/blob/5780093887cc092771c3922b1a4a965370cff510/test/lib.sh#L5-L6) and [this](https://github.com/openshift/boilerplate/blob/5780093887cc092771c3922b1a4a965370cff510/boilerplate/test/test-base-convention/update#L31-L32). You'll need to force-update and push the tag to the new commit when done.

cblecker:
@2uasimojo updated

2uasimojo (approved):
Thanks @cblecker 

/lgtm
/approve

PR #161, opened 2021-04-28
Support building multiple images

2uasimojo:
Previously the app-sre pipeline scripts were designed to build/push (if not already extant) exactly two things: the operator image and the OLM catalog.

With this commit we allow the consumer to build arbitrary images in addition to the operator image by setting a multiline `make` variable as follows:

```make
define ADDITIONAL_IMAGE_SPECS
./path/to/a/Dockerfile $(IMAGE_REGISTRY)/$(IMAGE_REPOSITORY)/a-image:v1.2.3
./path/to/b/Dockerfile $(IMAGE_REGISTRY)/$(IMAGE_REPOSITORY)/b-image:v4.5.6
endef
```

NOTE: As an ancillary side effect (benefit) of this commit, we stop tagging `latest` operator images. They were unused at best, misused at worst.

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/app-sre-build-deploy.sh
+++ b/boilerplate/openshift/golang-osd-operator/app-sre-build-deploy.sh
@@ -3,7 +3,23 @@
 set -ev
 
 usage() {
-    echo "$0 OPERATOR_IMAGE_URI REGISTRY_IMAGE CURRENT_COMMIT"
+    cat <<EOF
+    Usage: $0 REGISTRY_IMAGE CURRENT_COMMIT "IMAGE_SPECS"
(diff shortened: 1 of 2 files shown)


2uasimojo:
Reviewers: I tested this out with splunk-forwarder-operator by (removing the `build-push` override and) adding this to the Makefile:

```make
define ADDITIONAL_IMAGE_SPECS
# The splunk-forwarder image
$(FORWARDER_DOCKERFILE) $(FORWARDER_IMAGE_URI)
# The splunk-heavyforwarder image
$(HEAVYFORWARDER_DOCKERFILE) $(HEAVYFORWARDER_IMAGE_URI)
endef
```

PR #162, opened 2021-05-10
Split out stderr for `skopeo inspect`

2uasimojo:
This started happening suddenly:

```
10:08:43 parse error: Invalid literal at line 1, column 6
10:08:43 Unexpected error: skopeo inspect succeeded, but output contained no .Digest
10:08:43 Here's the output:
10:08:43 time="2021-05-10T15:08:41Z" level=error msg="HEADER map[Content-Length:[185] Content-Type:[text/html] Date:[Mon, 10 May 2021 15:08:41 GMT] Server:[nginx/1.12.1]]"
10:08:43 {
10:08:43     "Name": "quay.io/app-sre/aws-account-operator",
10:08:43     "Digest": "sha256:02baefe6abe17a60f9751441fdd7223b3d79af5580639d4ea90285bbab28a0c7",
```

For some reason, `skopeo inspect` is now producing that header information on stderr, whereas it wasn't before.

Nevertheless, that's no excuse for glomming together stdout and stderr when we expect to be able to parse the former. So with this commit, we split stderr out and redirect it to a temporary file instead.

### DIFF
--- a/boilerplate/_lib/common.sh
+++ b/boilerplate/_lib/common.sh
@@ -85,25 +85,34 @@ current_branch() {
 image_exists_in_repo() {
     local image_uri=$1
     local output
+    local rc
 
-    output=$(skopeo inspect docker://${image_uri} 2>&1)
-    if [[ $? -eq 0 ]]; then
+    local skopeo_stderr=$(mktemp)
+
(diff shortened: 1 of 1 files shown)


2uasimojo:
/hold

for testing (at least one bug exists)

iamkirkbater:
/lgtm

Looks sane to me.

PR #163, opened 2021-05-11
Secure freeze-check, enable hiding boilerplate deltas

2uasimojo:
- Document the "trust and ignore" philosophy that subscribers should adopt when reviewing PRs that include boilerplate updates.
- Secure `freeze-check` so you actually _can_ trust such changes.
- Lay down .gitattributes content such that, if a subscriber decides to suppress boilerplate deltas by default, certain files will still always be shown:
  - Because an attacker could attempt to subvert `freeze-check` by modifying that script or its dependencies, we explicitly exclude those from the hiding-by-default.
  - Because an attacker could try to subvert the hiding by appending entries to .gitattributes, we exclude THAT file from the hiding-by-default as well.

Security is hard.

### DIFF
--- a/boilerplate/update
+++ b/boilerplate/update
@@ -110,6 +110,12 @@ EOF
   echo "Updating the update script."
   rsync -a "${BP_CLONE}/boilerplate/update" "$0"
   echo "Copying utilities"
+  # HACK: Delete the utility dirs first because, in CI, rsync will fail
+  # to set things like mod times and permissions on the directories
+  # themselves, which have uid=gid=0.
+  for d in $(/bin/ls -d ${BP_CLONE}/boilerplate/_*); do
+    rm -fr $CONVENTION_ROOT/${d##*/}
+  done
   rsync -a -r --delete ${BP_CLONE}/boilerplate/_* $CONVENTION_ROOT
   echo "Reinvoking..."
   echo ""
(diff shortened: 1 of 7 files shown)


boilerplate/update
```diff
@@ -196,6 +202,35 @@ for convention in $(scrubbed_conventions $CONFIG_FILE); do
   echo ""
 done
 
+# (Create and) edit .gitattributes to
```
dustman9000, line 205: I'd really like to keep this as an optional config. Some consumers actually want to see the boilerplate diffs in their PRs.

2uasimojo: > I'd really like to keep this as an optional config. Some consumers actually want to see the boilerplate diffs in their PRs.

> You also have the potential of clobbering an existing .gitattributes.

We only clobber (with `sed`) and replace (by _appending_) the `GENERATED` section. It should be obvious that you shouldn't muck with that section, though I suppose I can add an extra comment in there. But as long as you don't, the only clobbering is caused by later directives overriding earlier ones. Which is 100% intentional because:

In order to instill confidence that the boilerplate changes in the PR don't deviate from what's in bp itself, we must also make sure that freeze-check itself hasn't been messed with. So we want to make sure it and its dependencies are revealed by default.

> Some consumers actually want to see the boilerplate diffs in their PRs.

Ah, so actually what we could do is leave the first directive (`boilerplate/** linguist-generated=true`) out of there, allowing consuming repos to decide whether they want it or not. If they do, the `GENERATED` section will still be applied after, to ensure certain deltas are revealed by default for the reasons described above. If they don't, the `GENERATED` section is a no-op.

Would that satisfy?

2uasimojo: I did these things.

2uasimojo:
/hold

for security review

jaybeeunix:
/lgtm

from a security perspective

2uasimojo:
/hold cancel

Let's do this.

PR #164, opened 2021-05-12
image-v1.0.1: Ratchet base image to 1.0.0

2uasimojo:
To make presumbit CI faster, this commit ratchets up the CI image build to "start" from the previously-released image, `image-v1.0.0`.

The original build.sh is renamed and a fresh build.sh is introduced. The latter is currently a no-op; subsequent image releases can add to it.

An app-sre-specific Dockerfile combines all the build scripts to do a full build from scratch in the appsre pipeline.

There is no functional change to the image itself.

[OSD-7253](https://issues.redhat.com/browse/OSD-7253)


config/build_image-v0.5.2.sh
```diff
@@ -0,0 +1,156 @@
+#!/usr/bin/env bash
```
2uasimojo, line 1: Note: This is an almost-100% rename of build.sh. The only thing different is the comment below.

config/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # NOTE: Keep this in sync with .ci-operator.yaml
-FROM registry.ci.openshift.org/openshift/release:golang-1.13
+FROM registry.ci.openshift.org/openshift/boilerplate:image-v0.5.2
```
cblecker, line 2: Is this a good idea? This means we won't get any security fixes if the underlying image is rebuilt

2uasimojo: By "underlying image" you mean `release:golang-1.13`? That's true. (To editorialize: relying on the moving-target behavior of a tagged image is just as bad as the producers of that image relying on us relying on that.)

If security (or other) fixes went into the base image, we wouldn't have picked them up anyway until the next time we cut a new version, because the [push script only uploads a new tagged version if we're on a tagged commit](https://github.com/openshift/boilerplate/blob/58c86db5160b06c2d0f3ee7e34f7ccc9ba582b2c/config/push.sh#L34-L38).

Which means deliberate action is required in order to pull in such fixes. Before this PR, a null image rev bump with a new tag. After this PR... something else, probably involving mucking with these files.

But even then, consuming repositories would have to merge a(ny) PR to pick up the changes. (Worth noting that the consumers only use the image for testing/linting/building the operator anyway, not for running it.)

That push script is kind of a mess anyway. It's building and pushing the image with `latest` and `$commit_hash` tags for every PR, even though those are a) duplicates of the most recent versioned image, and b) never used. We could fix it to
- stop doing that
- always rebuild the prior image(s), based on the renamed Dockerfile(s)
- always rebuild the most recent versioned image, which would then be based on ^

The point of this PR is to *not* always rebuild from scratch in prow, and I would really like to preserve that behavior. Having longer postsubmit build times is acceptable though. If we only modify the push script as described above, we achieve that -- but at the cost of the presubmit and postsubmit potentially building slightly different things. However, the only thing *this* repo runs against that image is shell code, so it's "never" going to notice the difference.

Thoughts?

2uasimojo: Okay, I did this.

jharrington22 (approved):
Nice. 

/lgtm

PR #165, opened 2021-05-14
Improve `subscriber report release`

2uasimojo:
This commit improves the `subscriber report release` tool in the following ways:
- It now requires a list of subscribers, or `ALL`, with the same semantics as `subscriber propose update`.
- It now produces a diff with which you can `patch -R` what's in the release repository to make it conform.


jharrington22 (approved):
/lgtm

PR #168, opened 2021-05-18
Adds .gitattributes to unacceptable_deltas "ignores" list

clcollins:
With a recent update to boilerplate, a .gitattributes file is created
during the bootstrap process.  This is causing failures in the
"framework/04-update-from-master-and-revert" test, which currently
doesn't expect the .gitattributes file to exist, uncommitted, after the
test bootstrap.

This adds the .gitattributes file to the "ignores" list in the
unacceptable_deltas function of boilerplate/update.

Also runs `./boilerplate/_lib/boilerplate-commit` to "clean" before the
next test run.

Signed-off-by: Christopher Collins <collins.christopher@gmail.com>

### DIFF
--- a/boilerplate/update
+++ b/boilerplate/update
@@ -73,12 +73,14 @@ scrubbed_conventions() {
 # This function ignores uncommitted changes to:
 # - Makefile: Editing this file is part of bootstrapping, and we don't
 #   want to force an additional commit in the bootstrapping process.
+# - .gitattributes: This file is created as part of the bootstrapping
+#   process. See above.
 # - ?? boilerplate/: I.e. the boilerplate subdirectory is brand new,
 #   meaning you're bootstrapping. See above.
 # - boilerplate/update.cfg: Changing this file prior to an update is
 #   part of the process of subscribing to new conventions.
 unacceptable_deltas() {
-  ignores="Makefile|boilerplate/(update\.cfg)?"
+  ignores="Makefile|.gitattributes|boilerplate/(update\.cfg)?"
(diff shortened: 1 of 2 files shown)


2uasimojo:
Okay, I think this is an appropriate thing to do, so +1 to the idea.

To make the test work, I suspect we may need to hack it to manually delete the `.gitattributes` file for the 'revert' step.

2uasimojo:
Played with this locally. This fixes the test:

```diff
diff --git a/test/case/framework/04-update-from-master-and-revert b/test/case/framework/04-update-from-master-and-revert
index dbfe13f..b469a52 100755
--- a/test/case/framework/04-update-from-master-and-revert
+++ b/test/case/framework/04-update-from-master-and-revert
@@ -35,6 +35,8 @@ override_boilerplate_repo $boilerplate_master || exit $?
 add_convention $repo test/test-base-convention
 make boilerplate-update || exit $?
 check_update $repo 04-check-after-init || exit $?
+# So the next update starts "clean"
+./boilerplate/_lib/boilerplate-commit
 
 reset_boilerplate_repo
 
```

I had [added that same thing further down](https://github.com/openshift/boilerplate/pull/163/files#diff-8beb88a246d3d85d2b5a3f7d23c299307a71f99222512adaa3de184dadeee798R43-R44) when I did the change that introduced .gitattributes; but since the change wasn't yet in master, it didn't blow up in that earlier phase until it was :)

2uasimojo:
/lgtm
/approve

Thanks for fixing this blocker, @clcollins !

PR #171, opened 2021-05-20
Use controller-gen for AAO

drpaneas:
AAO is separating its API as a sub-module so other projects can
import it in isolation without getting the whole operator.
To do this, the 'operator-sdk generate crds' need to be replace
with the `controller-gen`.

See https://github.com/openshift/aws-account-operator/pull/580
and https://issues.redhat.com/browse/OSD-7287


2uasimojo:
Definitely don't want to snowflake AAO from within boilerplate.

Happy to cut over to controller-gen fully (for everyone) once we can make it work compatibly. Reducing our dependency on operator-sdk 0.x is goodness.

PR #176, opened 2021-05-26
image-v2.0.0: Get rid of operator-sdk generate

2uasimojo:
Remove invocations of and support for `operator-sdk generate`, replacing these with the corresponding `controller-gen` calls in the `op-generate` target of openshift/golang-osd-operator's standard.mk.

The operator-sdk-generate.sh helper script is removed. We invoke controller-gen directly from the `make` target.

We need to preserve and add to the post-CRD-gen `yq` hacks to produce CRDs compatible with both v3 and v4. These should be able to go away once 3.11 is dead. (At that time we'll also need to flag controller-gen to produce CRD v1 instead of v1beta1.)

As written, this will work for consumers whether their APIs are packaged separately (as in e.g. openshift/aws-account-operator#580) or not.

With this commit, we produce a fresh backing image that omits the operator-sdk binaries, but is otherwise the same as image-v1.0.1.

With this commit, we're rebuilding the backing image from scratch, so the prow and jenkins Dockerfiles are the same. A subsequent commit will ratchet the former to be based on image-v2.0.0 to speed up builds, in a spirit similar to #164.

[OSD-7352](https://issues.redhat.com/browse/OSD-7352)
[OSD-7353](https://issues.redhat.com/browse/OSD-7353)


2uasimojo:
I tested the generated CRDs on v3 and v4. Found I had to remove type annotations [like these](https://github.com/openshift/aws-account-operator/pull/580/files#diff-d2f9264c85364eda036164f2005e67e2dd25f0701374e4d3d23742fc7f764adcR36-R38) to make it work.

This is ready to go now.

PR #182, opened 2021-06-03
podman enablement

2uasimojo:
- Podman and docker use different mechanisms to override the default path to the credentials cache. Accommodate both.
- Podman and docker use different transport prefixes to reference locally built images via skopeo. Accommodate both.
- Expand addition of `--userns keep-id` and `-v ...:Z` to more places.
- Resolve [OSD-6941](https://issues.redhat.com/browse/OSD-6941) by detecting the container engine in catalog-build.sh.

Co-Authored-By: @dofinn

### DIFF
--- a/boilerplate/test/test-base-convention/golint.sh
+++ b/boilerplate/test/test-base-convention/golint.sh
@@ -26,7 +26,7 @@ Exits zero if files are clean; nonzero if discrepancies are found.
 OPTIONS
     -f, --fix
         Fixes the problems it finds. (Don't forget to commit the
-	changes.)
+        changes.)
 
 TARGETS:
 $targets
@@ -49,12 +49,12 @@ eval set -- "$OPTS"
 FIX=
 while true; do
     case "$1" in
-	-f | --fix )
-	    FIX=1
-	    shift
-	    ;;
+        -f | --fix )
+            FIX=1
+            shift
+            ;;
         -h | --help ) usage;;
-	-- ) shift; break ;;
+        -- ) shift; break ;;
     esac
 done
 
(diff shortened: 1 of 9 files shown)


2uasimojo:
cc @dofinn @rporres

2uasimojo (commented):
Need to look closer, but the CI failure is happening because the test case is trying to run the podman/docker discovery and failing on both (because it's running in a container), which is killing the test under `-e`.

dofinn:
@2uasimojo  prob failing here https://github.com/openshift/boilerplate/pull/182/files#diff-ab5321e2d69263f527fe54b0c41c12438c8b0b4f2c9e121280c1f95cd091b5f1R46 not sure how we pass in the engine here tho as weneed it to determine the container transport that skopeo uses.. also testing locally with `ALLOW_DIRTY_CHECKOUT=true make pr-test` does not weild the exact same results as this.

2uasimojo:
@dofinn I fixed the test, rebased, squashed, and tidied up a couple more things. Please validate this in your env and unhold if it works. I guess you could also lgtm if you like -- we're kind of validating each other's pieces -- or if you can find someone else to review.

boilerplate/test/test-base-convention/golint.sh
```diff
@@ -26,7 +26,7 @@ Exits zero if files are clean; nonzero if discrepancies are found.
 OPTIONS
     -f, --fix
         Fixes the problems it finds. (Don't forget to commit the
-	changes.)
+        changes.)
```
2uasimojo, line 29: Unrelated whitespace fixups. Sue me.

2uasimojo:
/hold cancel

@bdematte @dustman9000 or @jharrington22 would one of you please review and lgtm this?

PR #183, opened 2021-06-04
Skip update/revert test case if at master

2uasimojo:
If we're already at master (which happens e.g. during rehearsals on CI config updates), the 04-update-from-master-and-revert test is a) silly, and b) going to fail when attempting `boilerplate-commit` because there's nothing to update and therefore nothing to commit. Add logic to short out of this test case in this scenario.

### DIFF
--- a/test/case/framework/04-update-from-master-and-revert
+++ b/test/case/framework/04-update-from-master-and-revert
@@ -7,6 +7,16 @@ Tests updating from the current boilerplate master to the current
 commit, and then reverting to the master.
 EOF
 
+# Sometimes, e.g. when openshift/release runs this test in a rehearsal,
+# we're *at* master. In such cases, this test would legitimately fail on
+# `boilerplate-commit` because there's nothing to update and therefore
+# nothing to commit. We can skip the test in such cases (master has
+# already been tested anyway).
+if [[ $(git rev-parse HEAD) == $(git rev-parse master) ]]; then
+    echo "==SKIP== We're already at master"
+    exit 0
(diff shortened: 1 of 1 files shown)


2uasimojo:
For reviewers: Here's what the test case does at `master`:

```
bash-4.2$ ALLOW_DIRTY_CHECKOUT=true make CASE_GLOB=04-update-from-master-and-revert test
config/tag-check.sh
++ git rev-list --no-merges -n 1 HEAD
+ commit=7ac2a583973b4799c685d701d784624bb815b729
++ git describe --exact-match --tag --match 'image-v*' 7ac2a583973b4799c685d701d784624bb815b729
++ true
+ tag=
+ [[ -n '' ]]
++ git log --merges -n1 'HEAD^' --format=%H
+ last_merge_commit=c6d84f8c6e1e786db9d0ced72def1e7f3a7bb124
++ git diff c6d84f8c6e1e786db9d0ced72def1e7f3a7bb124 --name-only config/
+ [[ -n '' ]]
+ exit 0
test/driver 04-update-from-master-and-revert
Will run test cases:
./framework/04-update-from-master-and-revert

=========================
Running test case framework/04-update-from-master-and-revert
=========================
Tests updating from the current boilerplate master to the current
commit, and then reverting to the master.
==SKIP== We're already at master
=========================
PASS

=========================
PASS: 1
  framework/04-update-from-master-and-revert
FAIL: 0
  
=========================

Cleaning up
Removing temporary directories
```

PR #184, opened 2021-06-14
Remove efried per Leaver process

billmvt:
see https://issues.redhat.com/browse/OHSS-4669


billmvt:
Rejecting this change. Decided with @jharrington22 to leave Eric as Approver for Boilerplate.

PR #185, opened 2021-06-15
Remove codecov secret mapping stuff

2uasimojo:
This is no longer applicable since the move to self-service vault.


jharrington22 (approved):
/lgtm

PR #189, opened 2021-08-09
Add OWNERS_ALIASES file to all repos

cblecker:


2uasimojo:
IIUC, this will allow you to add `srep-team-leads` to the `OWNERS` file in each individual repo, thereby authorizing the members thereof to approve PRs therein?

If that's the case, perhaps we would also like to add a yq ensuring that that alias is present in the repo's OWNERS. And document it in the convention's [README](https://github.com/openshift/boilerplate/blob/b177042403389885872d904a003e3adcf6350655/boilerplate/openshift/golang-osd-operator/README.md).

PR #195, opened 2021-11-04
golang-osd-operator appsre: Fix docker login

2uasimojo:
When stealing (ahem, "reappropriating") podman/docker-accommodating
App-SRE pipeline code when hive switched to a rhel8 jenkins node, we
found a latent bug on the docker side: the `REGISTRY_AUTH_FILE` variable
needs to point to an actual file, not just the directory in which the
file is located. Contributing the fix back "upstream".


2uasimojo:
/assign @dofinn 

Dom, would you mind testing this in your podman-havin env? `make REGISTRY_USER=xxx REGISTRY_TOKEN=yyy docker-login` should fail without this fix, succeed with it.

2uasimojo:
Actually I'm needing this fix to test appsre pipelines locally too (still on rhel8 with real docker). Not sure what changed.

PR #198, opened 2021-11-26
App-SRE testing docs should suggest to rebase saas fork

mrbarge:
This PR amends the "testing app-sre" documentation to recommend that the personal `saas-<operator>-bundle` fork be rebased off upstream if it already exists. Not doing this step has occasionally caused me issues with a badly operating catalog when I've previously polluted my fork's staging/production branches from prior tests.


2uasimojo (commented):
Or "remove and recreate your fork" -- particularly since it isn't necessary to clone the fork locally for testing to work.

But yeah, love this.

/lgtm

PR #199, opened 2021-11-29
Update catalog-build.sh

georgettica:
Based on https://github.com/georgettica/route-monitor-operator/pull/4
Which is automated

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/csv-generate/catalog-build.sh
+++ b/boilerplate/openshift/golang-osd-operator/csv-generate/catalog-build.sh
@@ -63,7 +63,7 @@ EOF
 
 # Build registry
 cat <<EOF > $DOCKERFILE_REGISTRY
-FROM quay.io/openshift/origin-operator-registry:4.8.0
+FROM quay.io/openshift/origin-operator-registry:4.10.0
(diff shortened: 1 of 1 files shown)


2uasimojo:
This seems like a reasonable thing to do, but I'd like to see some evidence that it works first. ISTR we had compat problems with 4.9, which is why we didn't move to it right away.

georgettica:
how do we run the testing?
is there a dry run I can invoke?

2uasimojo:
> how do we run the testing? is there a dry run I can invoke?

You can pull this change into a fully onboarded OSD operator (so not RMO) and run it through [this process](https://github.com/openshift/boilerplate/blob/master/boilerplate/openshift/golang-osd-operator/app-sre.md). That will build (operator and) catalog images and push them to your personal quay repo. From there you can install them in a scratch cluster. (Obviously to test this particular change you will need to install from the catalog via OLM -- just swapping out the operator image is not sufficient.)

ravitri:
@georgettica - A recent change involving PR https://github.com/openshift/boilerplate/pull/218 updated this image to 4.10 as well along with rebuilding the registry image using ubi-micro base image. Thus, we could close this PR in favor of it.

PR #202, opened 2022-02-06
Support empty SaaS repositories

mrbarge:
This PR updates boilerplate's `golang-osd-operator` to support initializing empty SaaS repositories.

Presently, the CSV and bundle generation scripts have a dependency on there being:
- An existing, populated SaaS repository
- An existing 'previous' CSV version that the current CSV will mark as replacing.

When onboarding a new operator, neither of these things will be true [0]. This PR updates the script to add the following behaviours:

- If the git cloned SaaS repository is empty, create the subdirectory in which the bundles will reside.
- If a previous CSV version cannot be detected, do not pass it to the bundle generator.
- The bundle generator will not add a `replaces` clause in the CSV if there is no previous version. (invalid `replaces` entries  result in a broken catalog that will not respond to grpc package queries)

Successfully tested on the following scenarios:
- A fresh, empty SaaS repository.
- An already established SaaS repository (to verify no breaking changes)

[0] Presently this is blocking SaaS onboarding of the `ocm-agent-operator`.


boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
```diff
@@ -93,6 +93,18 @@ else
     rm -rf "$SAAS_OPERATOR_DIR"
     git clone --branch "$operator_channel" ${GIT_PATH} "$SAAS_OPERATOR_DIR"
 
+    # First exit appropriately if the clone could not succeed
+    if [[ $? != 0 ]]; then
```
2uasimojo, line 97: Since this script is running under `set -e` (L3), I would expect it to exit right away if the `git clone` on L94 fails.

2uasimojo (approved):
Legit, thanks Matt.

One redundant (never reached) check noted inline. Not a blocker -- remove if you like, or unhold if you want to ignore it.

/lgtm
/hold

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
mrbarge: Thanks a lot for the review @2uasimojo ! And thanks for the catch - I've removed that superfluous check now.

PR #203, opened 2022-02-15
fix(CONTAINER_ENGINE): allow setting the env

georgettica:
after #201 didn't work, I updated it a bit

now you can set the env and it's not overriden


2uasimojo:
Okay.

/lgtm

PR #205, opened 2022-03-23
OSD-10311 - Initial commit for golang-codecov and golang-lint conventions

bdematte:
In order to be able to use boilerplate in non-operators (like for osd-network-verifier in [OSD-10311](https://issues.redhat.com/browse/OSD-10311) ), we need to split the existing operator convention to be able to use pieces in a standalone fashion.
Agreed approach for this is to first create new conventions for those smaller chunks, and then have a dedicated task to clean-up and refactor the existing operator convention ( opened https://issues.redhat.com/browse/OSD-10468 for the 2nd step ). 
Minutes from initial discussion : https://docs.google.com/document/d/1CbMiQLVrUn-3B88-P_9v8Wv560-ystAo1yBU0PW4FFs/edit#

### DIFF
--- a/boilerplate/openshift/golang-codecov/README.md
+++ b/boilerplate/openshift/golang-codecov/README.md
@@ -4,10 +4,10 @@
 
 ### Prow
 
-| Test name / `make` target | Purpose                                                                                                             |
-|---------------------------|-----------------------------------------------------------------------------------------------------------------    |
-| `test`                    | "Local" unit and functional testing.                                                                                |
-| `coverage`                | [Code coverage](#code-coverage) analysis and reporting.                                                             |
+| `make` target | Purpose                                                                                                             |
+|---------------|-----------------------------------------------------------------------------------------------------------------    |
+| `test`        | "Local" unit and functional testing.                                                                                |
+| `coverage`    | [Code coverage](#code-coverage) analysis and reporting.                                                             |
 
 
 ## Code coverage
--- a/boilerplate/openshift/golang-lint/README.md
+++ b/boilerplate/openshift/golang-lint/README.md
@@ -2,9 +2,9 @@
 
 - [Linting and other static analysis with `golangci-lint`](#linting-and-other-static-analysis-with-golangci-lint)
 
-| Test name / `make` target | Purpose                                                                                                             |
-|---------------------------|-----------------------------------------------------------------------------------------------------------------    |
-| `lint`                    | Perform static analysis.                                                                                            |
+| `make` target | Purpose                                                                                                             |
+|---------------|-----------------------------------------------------------------------------------------------------------------    |
+| `lint`        | Perform static analysis.                                                                                            |
 
 
 ## Linting and other static analysis with `golangci-lint`


boilerplate/openshift/golang-lint/ensure.sh
```diff
@@ -0,0 +1,88 @@
...
+        if [ $IN_PATH != 0 ]; then
+            echo "${GOPATH}/bin not in $$PATH"
+            exit 1
+        fi
+        DOWNLOAD_URL="https://github.com/golangci/golangci-lint/releases/download/v${GOLANGCI_LINT_VERSION}/golangci-lint-${GOLANGCI_LINT_VERSION}-${GOOS}-amd64.tar.gz"
+        curl -sfL "${DOWNLOAD_URL}" | tar -C "${GOPATH}/bin" -zx --strip-components=1 "golangci-lint-${GOLANGCI_LINT_VERSION}-${GOOS}-amd64/golangci-lint"
+    fi
+    ;;
+
+opm)
```
bdematte, line 33: Should be cleaned up as not needed in the scope of `lint` only

boilerplate/openshift/golang-codecov/standard.mk
```diff
@@ -0,0 +1,24 @@
...
+TESTOPTS :=
+
+# coverage: Code coverage analysis and reporting.
+.PHONY: coverage
+coverage:
+	${CODECOV_CONVENTION_DIR}/codecov.sh
+
+.PHONY: go-test
+go-test:
+	${GOENV} go test $(TESTOPTS) $(TESTTARGETS)
```
2uasimojo, line 24: We ought to think ahead to when this convention will be one of many imported by an operator, and `go-test` may be defined in more than one of them. Today in golang-osd-operator it's [aliased](https://github.com/openshift/boilerplate/blob/7d81a9d4ba6ed1b17ed0f0ceee85cb9def9884d4/boilerplate/openshift/golang-osd-operator/standard.mk#L242-L244) to `test`, which is used for CI, and it's one of the targets run by [default](https://github.com/openshift/boilerplate/blob/7d81a9d4ba6ed1b17ed0f0ceee85cb9def9884d4/boilerplate/openshift/golang-osd-operator/standard.mk#L105) (argless `make`). Perhaps we should rename this one `codecov-go-test` to avoid conflicts.

For that matter, we need to think about the source of truth for shared variables like `GOENV`, `TESTOPTS`, and `TESTTARGETS`.

As currently implemented, if you consume multiple conventions, rules/vars will take [precedence](https://github.com/openshift/boilerplate/blob/7d81a9d4ba6ed1b17ed0f0ceee85cb9def9884d4/boilerplate/update#L201) in the reverse of the [order](https://github.com/openshift/boilerplate/blob/7d81a9d4ba6ed1b17ed0f0ceee85cb9def9884d4/boilerplate/update#L173) you list them in your [config](https://github.com/openshift/boilerplate/blob/7d81a9d4ba6ed1b17ed0f0ceee85cb9def9884d4/boilerplate/update#L34). So unless we do something else/different, it's on the consumer to list conventions in ascending order of importance.

boilerplate/openshift/golang-codecov/update
```diff
@@ -0,0 +1,36 @@
...
+include boilerplate/generated-includes.mk
+
+- Remove any other 'include' lines, unless they're for things truly
+  unique to your repository. (Otherwise, consider proposing them to
+  boilerplate.)
+
+- Delete any obsolete files you're no longer including.
+
+- Make sure you are properly integrated with codecov.io:
+  https://github.com/openshift/ops-sop/blob/93d100347746ce04ad552591136818f82043c648/services/codecov.md#generate-the-codecovio-token
```
2uasimojo, line 32:
```suggestion
  https://github.com/openshift/ops-sop/blob/8e48d0c1e8d9f2f5a19b1e18cdf5f08f658eb184/services/codecov.md#generate-the-codecovio-token
```

boilerplate/openshift/golang-codecov/README.md
```diff
@@ -0,0 +1,14 @@
+# Conventions for code coverage for our Golang projects
+
+- [Code coverage](#code-coverage)
+
```
2uasimojo, line 4: This README could mention the `make` targets that are also included, and how this convention ties into the `coverage` CI target.

On that note, should we include a `prow-config` script that lays down just the codecov subset of the [one from golang-osd-operator](https://github.com/openshift/boilerplate/blob/7d81a9d4ba6ed1b17ed0f0ceee85cb9def9884d4/boilerplate/openshift/golang-osd-operator/prow-config)?

Looking to the future, assembling piecemeal conventions like this, each convention's prow-config will have to be more clever, actually parsing the yaml and ensuring its specific section(s) are copacetic rather than laying down the whole file.

boilerplate/openshift/golang-lint/ensure.sh
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env bash
```
2uasimojo, line 1: Currently an exact copy of the one from golang-osd-operator -- but agree with comment below that it should be stripped down to just the lint-related bits.

boilerplate/openshift/golang-lint/golangci.yml
```diff
@@ -0,0 +1,28 @@
+# options for analysis running
```
2uasimojo, line 1: Identical to the one in golang-osd-operator ✓

boilerplate/openshift/golang-lint/standard.mk
```diff
@@ -0,0 +1,16 @@
...
+
+.PHONY: go-check
+go-check: ## Golang linting and other static analysis
+	${LINT_CONVENTION_DIR}/ensure.sh golangci-lint
+	GOLANGCI_LINT_CACHE=${GOLANGCI_LINT_CACHE} golangci-lint run -c ${LINT_CONVENTION_DIR}/golangci.yml ./...
+	test "${GOLANGCI_OPTIONAL_CONFIG}" = "" || test ! -e "${GOLANGCI_OPTIONAL_CONFIG}" || GOLANGCI_LINT_CACHE="${GOLANGCI_LINT_CACHE}" golangci-lint run -c "${GOLANGCI_OPTIONAL_CONFIG}" ./...
+
+# lint: Perform static analysis.
+.PHONY: lint
+lint: go-check
```
2uasimojo, line 16: So yeah, again here, golang-osd-operator includes another target in here. That ought to shake out fine as long as it's listed later.

boilerplate/openshift/golang-codecov/standard.mk
bdematte: Yes, those are points that I added to the docs mentioned in the description as this needs to be looked at when finalizing the split (ie removing the dupps from the operator convention, to be done in [OSD-10468](https://issues.redhat.com/browse/OSD-10468) . 
For the very short term, variables are defined the same way in all conventions and we should finish the migration (answering the 'how to') before it becomes troublesome.

boilerplate/openshift/golang-codecov/.codecov.yml
```diff
@@ -0,0 +1,26 @@
+codecov:
```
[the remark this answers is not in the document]
bdematte: The objective of this first step is to make a simple/standalone convention, hence the copied files. Once strategy for golang-osd-operator is defined, it will cover the clean-up of those dupps ( [OSD-10468](https://issues.redhat.com/browse/OSD-10468) )

2uasimojo:
Thanks for the updates @bdematte 

Is someone from SD going to be your official reviewer/approver?

bdematte:
I will try to get review from James or Dustin to ensure this is in line with discussions and I didn't miss/forget something. Thanks for your review @2uasimojo

2uasimojo:
/lgtm

Low risk as this is a new convention and won't affect existing consumers. This needs some work, as noted inline, but this is a good start and can be iterated on.

jharrington22:
Thanks @2uasimojo for the reviews I read through the comments and lgtm too @bdematte

PR #206, opened 2022-03-23
feat(golangci): add gosec

georgettica:
per ticket  OSD-10161, this change should be running via CI on all osd operators.

this way we are always compliant to the gosec (as we have done one audit a forever ago and cleaned issues

this change should get approval before merging as this might cause initial work to upgrade to this version of boilerplate

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/golangci.yml
+++ b/boilerplate/openshift/golang-osd-operator/golangci.yml
@@ -18,6 +18,7 @@ linters:
   enable:
   - deadcode
   - errcheck
+  - gosec
   - gosimple
   - govet
   - ineffassign


2uasimojo:
Reasonable.

If you want to preview how this affects CI for a consuming repo before committing to it, you can set Makefile variable `GOLANG_CI_OPTIONAL_CONFIG` ([ref](https://github.com/openshift/boilerplate/blob/b65f892255725d9a566f5aaf50b8c34dd3b38381/boilerplate/openshift/golang-osd-operator/standard.mk#L149)) to point to a local file containing something like:

```yaml
linters:
  disable-all: true
  enable:
  - gosec
```

georgettica:
Oki!
Then let's merge and add this to repo maintainers to adopt

2uasimojo:
> let's merge and add this to repo maintainers to adopt

I don't understand. Merge what?

2uasimojo:
Sorry, what I mean is that owners of _consuming_ repos can get an understanding for how this boilerplate PR will affect them when they pull it in -- and perhaps they should do that before we merge this just in case there are show-stopping issues that will take them some time to resolve.

2uasimojo:
Like [this](https://github.com/openshift/aws-account-operator/pull/685). You can see this will cause `lint` to fail for a-a-o. Proposing (if not merging) a similar PR across all the OSD repositories would be pretty easy, and would alert you in case any of the problems it uncovers end up being really hard to fix. That way you don't get stuck not being able to upgrade boilerplate _past_ this commit in case something more important is merged later.

georgettica:
I agree, and I can do something like that to all the repos so once the checks are fixed, I can promote to boilerplate

georgettica:
/close
as I lost interest :(

georgettica:
/reopen
seems people are talking on it, giving a second try 🤞

georgettica:
related to OSD-11325

georgettica:
I am for merging, and I am not sure I should force merge this.
Maybe someone who is approver can push this forward?

2uasimojo:
I'll
/approve
but
/hold
to make sure you've properly considered https://github.com/openshift/boilerplate/pull/206#issuecomment-1077797612
TL;DR: if you merge this, consuming repos will be forced to resolve any resulting errors before they can get a bp at *or after* this commit. So like, if you urgently need to pick up an unrelated bp update, you'll be stuck.

boilerplate/openshift/golang-osd-operator/golangci.yml
```diff
@@ -18,6 +18,7 @@ linters:
   enable:
   - deadcode
   - errcheck
+  - gosec
```
dustman9000, line 21: We should look at adding rules incrementally to address Eric's concern about all consuming repos failing to pass the lint PR checks once they consume this change. https://golangci-lint.run/usage/linters/#gosec

dustman9000: Maybe just G101-G112 for now? https://github.com/securego/gosec#available-rules

georgettica: Sure, feel free to recreate the PR.

Also I know @Tessg22 is doing the painstakingly hard process of checking per-repo.

Please collab with her if possible on this

Tessg22:
Since [Preview Golangci gosec lint in OSD operators](https://issues.redhat.com/browse/SDE-2056) is complete, we should be ready to unblock this PR

NautiluX:
/hold cancel
All operators cleared issues, this can be merged.

PR #208, opened 2022-03-25
Update README for image tagging quirks

2uasimojo:
Clarify a couple of things in the README wrt image tagging:
- You have to push the new tag to your `origin` as well as `upstream`.
- You have to edit the tag in a couple of in-repo files as well.


2uasimojo:
/assign @cblecker 

...unless you know a way to make tag-check.sh look only at the upstream.

PR #204, opened 2022-03-18
Switch to RHEL8 builder image

cblecker:
Switches and updates the base builder image to the OCP sanctioned RHEL8 image

PR #211, opened 2022-04-26
Update boilerplate to support latest osdk

wanghaoran1988:
This makes boilerplate support latest osdk.

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/standard.mk
+++ b/boilerplate/openshift/golang-osd-operator/standard.mk
@@ -46,6 +46,9 @@ OPERATOR_IMAGE_URI=${IMG}
 OPERATOR_IMAGE_URI_LATEST=$(IMAGE_REGISTRY)/$(IMAGE_REPOSITORY)/$(IMAGE_NAME):latest
 OPERATOR_DOCKERFILE ?=build/Dockerfile
 REGISTRY_IMAGE=$(IMAGE_REGISTRY)/$(IMAGE_REPOSITORY)/$(IMAGE_NAME)-registry
+#The api dir that latest osdk generated
+NEW_API_DIR=./api
+USE_OLD_SDK=$(shell if [[ -d "$(NEW_API_DIR)" ]];then echo FALSE;else echo TRUE;fi)
 
 # Consumer can optionally define ADDITIONAL_IMAGE_SPECS like:
 #     define ADDITIONAL_IMAGE_SPECS
@@ -67,7 +70,12 @@ REGISTRY_USER ?=
 REGISTRY_TOKEN ?=
 
 BINFILE=build/_output/bin/$(OPERATOR_NAME)
-MAINPACKAGE ?= ./cmd/manager
+MAINPACKAGE = ./main.go
+API_DIR = $(NEW_API_DIR)
+ifeq ($(USE_OLD_SDK), TRUE)
+MAINPACKAGE = ./cmd/manager
+API_DIR = ./pkg/apis
+endif
 
 GOOS?=$(shell go env GOOS)
 GOARCH?=$(shell go env GOARCH)
@@ -159,11 +167,32 @@ go-generate:
 	${GOENV} go generate $(TESTTARGETS)
 	# Don't forget to commit generated files
 
+# go-get-tool will 'go install' any package $2 and install it to $1.
+define go-get-tool
+@{ \
+set -e ;\
+TMP_DIR=$$(mktemp -d) ;\
+cd $$TMP_DIR ;\
+go mod init tmp ;\
+echo "Downloading $(2)" ;\
+GOBIN=$(shell dirname $(1)) go install $(2) ;\
+echo "Installed in $(1)" ;\
+rm -rf $$TMP_DIR ;\
+}
+endef
+
+CONTROLLER_GEN_VERSION = v0.8.0
+CONTROLLER_GEN = controller-gen-$(CONTROLLER_GEN_VERSION)
+ifeq ($(USE_OLD_SDK), TRUE)
+#If we are using the old osdk, we use the default controller-gen and it's v0.3.0 for now.
+CONTROLLER_GEN = controller-gen
+endif
+
+
 .PHONY: op-generate
 op-generate:
-	# The artist formerly known as `operator-sdk generate crds`:
+	cd $(API_DIR); $(CONTROLLER_GEN) crd paths=./... output:dir=$(PWD)/deploy/crds
 ifeq ($(CRD_VERSION), v1beta1)
-	cd pkg/apis; controller-gen crd paths=./... output:dir=../../deploy/crds
 	# HACK: Due to an OLM bug in 3.11, we need to remove the
 	# spec.validation.openAPIV3Schema.type from CRDs. Remove once
 	# 3.11 is no longer supported.
@@ -176,16 +205,17 @@ ifeq ($(CRD_VERSION), v1beta1)
 	find deploy/crds -name '*.yaml' | xargs -n1 -I{} yq d -i {} 'spec.**.x-kubernetes-list-type'
 	find deploy/crds -name '*.yaml' | xargs -n1 -I{} yq d -i {} 'spec.**.x-kubernetes-map-type'
 	find deploy/crds -name '*.yaml' | xargs -n1 -I{} yq d -i {} 'spec.**.x-kubernetes-struct-type'
-else
-	cd pkg/apis; controller-gen crd:crdVersions=v1 paths=./... output:dir=../../deploy/crds
 endif
-	# The artist formerly known as `operator-sdk generate k8s`:
-	cd pkg/apis; controller-gen object paths=./...
-	# Don't forget to commit generated files
+	cd $(API_DIR); $(CONTROLLER_GEN) object paths=./...
+
+API_DIR_MIN_DEPTH = 1
+ifeq ($(USE_OLD_SDK), TRUE)
+API_DIR_MIN_DEPTH = 2
+endif
 
 .PHONY: openapi-generate
 openapi-generate:
-	find ./pkg/apis/ -maxdepth 2 -mindepth 2 -type d | xargs -t -n1 -I% \
+	find $(API_DIR) -maxdepth 2 -mindepth $(API_DIR_MIN_DEPTH) -type d | xargs -t -I% \
 		openapi-gen --logtostderr=true \
 			-i % \
 			-o "" \
@@ -193,7 +223,7 @@ openapi-generate:
 			-p % \
 			-h /dev/null \
 			-r "-"
-
+	
 .PHONY: generate
 generate: op-generate go-generate openapi-generate
 
@@ -207,9 +237,24 @@ go-build: ## Build binary
 	# This is temporary until a better container build method is developed
 	${GOENV} GOOS=linux go build ${GOBUILDFLAGS} -o ${BINFILE} ${MAINPACKAGE}
 
+# ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
+ENVTEST_K8S_VERSION = 1.23
+
+SETUP_ENVTEST = $(shell pwd)/bin/setup-envtest
+.PHONY: setup-envtest
+setup-envtest: ## Download setup-envtest locally if necessary.
+	$(call go-get-tool,$(SETUP_ENVTEST),sigs.k8s.io/controller-runtime/tools/setup-envtest@latest)
+	$(eval KUBEBUILDER_ASSETS = "$(shell $(SETUP_ENVTEST) use $(ENVTEST_K8S_VERSION) -p path --bin-dir $(PWD)/bin)")
+	
+# Setting SHELL to bash allows bash commands to be executed by recipes.
+# This is a requirement for 'setup-envtest.sh' in the test target.
+# Options are set to exit when a recipe line exits non-zero or a piped command fails.
+SHELL = /usr/bin/env bash -o pipefail
+.SHELLFLAGS = -ec
+
 .PHONY: go-test
-go-test:
-	${GOENV} go test $(TESTOPTS) $(TESTTARGETS)
+go-test: setup-envtest
+	KUBEBUILDER_ASSETS=$(KUBEBUILDER_ASSETS) go test $(TESTOPTS) $(TESTTARGETS)
 
 .PHONY: python-venv
 python-venv:
--- a/boilerplate/openshift/golang-osd-operator/update
+++ b/boilerplate/openshift/golang-osd-operator/update
@@ -14,8 +14,9 @@ echo "Copying .codecov.yml to your repository root."
 cp ${HERE}/.codecov.yml $REPO_ROOT
 
 # TODO: boilerplate more of Dockerfile
-echo "Overwriting build/Dockerfile's initial FROM with $IMAGE_PULL_PATH"
-${SED?} -i "1s,.*,FROM $IMAGE_PULL_PATH AS builder," build/Dockerfile
+DOCKERFILE=build/Dockerfile
+echo "Overwriting $DOCKERFILE's initial FROM with $IMAGE_PULL_PATH"
+${SED?} -i "1s,.*,FROM $IMAGE_PULL_PATH AS builder," $DOCKERFILE
 
 echo "Writing .ci-operator.yaml in your repository root with:"
 echo "    namespace: $IMAGE_NAMESPACE"


boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -191,15 +209,30 @@ openapi-generate:
...
 
+
 .PHONY: go-build
 go-build: ## Build binary
 	# Force GOOS=linux as we may want to build containers in other *nix-like systems (ie darwin).
 	# This is temporary until a better container build method is developed
 	${GOENV} GOOS=linux go build ${GOBUILDFLAGS} -o ${BINFILE} ${MAINPACKAGE}
 
+# ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
+ENVTEST_K8S_VERSION = 1.23
```
ravitri, line 220: Is this version supposed to be the latest GA version that's used in OSD/ROSA clusters? If yes, should we include it in the gap analysis process [SOP](https://github.com/openshift/ops-sop/blob/master/v4/howto/gap-analysis.asciidoc)?

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -153,11 +153,30 @@ go-generate:
...
 
+# go-get-tool will 'go install' any package $2 and install it to $1.
+define go-get-tool
+@[ -f $(1) ] || { \
+set -e ;\
+TMP_DIR=$$(mktemp -d) ;\
+cd $$TMP_DIR ;\
+go mod init tmp ;\
+echo "Downloading $(2)" ;\
+GOBIN=${PWD}/bin go install $(2) ;\
```
2uasimojo, line 164: Looks like this will ignore $1 and install into $TMP_DIR/bin. Are we missing a step to copy it out to $1?

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -191,15 +209,30 @@ openapi-generate:
...
+# This is a requirement for 'setup-envtest.sh' in the test target.
+# Options are set to exit when a recipe line exits non-zero or a piped command fails.
+SHELL = /usr/bin/env bash -o pipefail
+.SHELLFLAGS = -ec
+
 .PHONY: go-test
-go-test:
-	${GOENV} go test $(TESTOPTS) $(TESTTARGETS)
+go-test: envtest
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test ./... -coverprofile cover.out
```
2uasimojo, line 235: I don't think this should be using `-coverprofile`. Leave that for the [`coverage` target](https://github.com/openshift/boilerplate/pull/211/files#diff-aef0082c5bb612084a57562263d822bb2a2a089ec35cfcf3feec179673c56b11R280), which operates via [codecov.sh](https://github.com/openshift/boilerplate/blob/bea9ec5ce249a5a7182598bbb4fea5ca148bbc5f/boilerplate/openshift/golang-osd-operator/codecov.sh) invoking `make go-test` with certain TESTOPTS.

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -67,7 +67,7 @@ REGISTRY_USER ?=
 REGISTRY_TOKEN ?=
 
 BINFILE=build/_output/bin/$(OPERATOR_NAME)
-MAINPACKAGE ?= ./cmd/manager
+MAINPACKAGE ?= ./main.go
```
2uasimojo, line 70: Does this mean consuming repos will need to move their main.go and/or override this variable when they update to/past this level of boilerplate?

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -171,15 +190,14 @@ ifeq ($(CRD_VERSION), v1beta1)
...
-	# The artist formerly known as `operator-sdk generate k8s`:
-	cd pkg/apis; controller-gen object paths=./...
-	# Don't forget to commit generated files
+	$(CONTROLLER_GEN) object paths=./...
+
 
 .PHONY: openapi-generate
 openapi-generate:
-	find ./pkg/apis/ -maxdepth 2 -mindepth 2 -type d | xargs -t -n1 -I% \
+	find ./api/ -maxdepth 2 -mindepth 2 -type d | xargs -t -n1 -I% \
```
2uasimojo, line 200: Similarly, does this mean consuming repos will have to move their API packages?

wanghaoran1988: This is true, without this, you cannot use `operator-sdk create api` to add new api.

boilerplate/openshift/golang-osd-operator/standard.mk
wanghaoran1988: yes, that's the plan so far, we are working on an step by step guide how to do the migration at the same time. 
I am also working on make boilerplate compatible by checking which osdk you are using.

boilerplate/openshift/golang-osd-operator/standard.mk
wanghaoran1988: thanks, this was copy&past from the Makefile generated by latest osdk.

boilerplate/openshift/golang-osd-operator/standard.mk
wanghaoran1988: > Is this version supposed to be the latest GA version that's used in OSD/ROSA clusters?

I am not quite sure, this version is used by latest osdk.

>If yes, should we include process of updating boilerplate image with new version as per gap analysis [SOP](https://github.com/openshift/ops-sop/blob/master/v4/howto/gap-analysis.asciidoc)?

what do you mean by this?

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -171,14 +202,14 @@ ifeq ($(CRD_VERSION), v1beta1)
...
-	# The artist formerly known as `operator-sdk generate k8s`:
-	cd pkg/apis; controller-gen object paths=./...
-	# Don't forget to commit generated files
+	cd $(API_DIR); $(CONTROLLER_GEN) object paths=./...
+
 
 .PHONY: openapi-generate
 openapi-generate:
+ifeq ($(USE_OLD_SDK), TRUE)
 	find ./pkg/apis/ -maxdepth 2 -mindepth 2 -type d | xargs -t -n1 -I% \
```
2uasimojo, line 213:
```suggestion
	find $(API_DIR) -maxdepth 2 -mindepth 2 -type d | xargs -t -n1 -I% \
```

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -187,19 +218,46 @@ openapi-generate:
...
+else
+	find ./api -maxdepth 2 -mindepth 1 -type d | xargs -t -n1 -I% \
+		openapi-gen --logtostderr=true \
+			-i % \
+			-o "" \
+			-O zz_generated.openapi \
+			-p % \
+			-h /dev/null \
+			-r "-"
+endif
```
2uasimojo, line 230: At a glance, these look to be the same except for the directory (already has a variable, as noted above) and the mindepth. If you added a variable for the latter, you could dedup this.

boilerplate/openshift/golang-osd-operator/standard.mk
wanghaoran1988: updated

boilerplate/openshift/golang-osd-operator/standard.mk
ravitri: Well, I just meant to ask about the frequency when this value `ENVTEST_K8S_VERSION = 1.23` will be updated. Right now, 4.10 (latest GA version) uses Kubernetes 1.23. So if this value is to be updated when 4.11 is GA, we might want to just track the change as part of gap analysis process in evaluating new version as well. Not sure if possible to have it done from a separate branch of boilerplate than `master` branch. Nothing majorly concerning right now so feel free to skip my query too.

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -187,19 +218,46 @@ openapi-generate:
...
 	# Force GOOS=linux as we may want to build containers in other *nix-like systems (ie darwin).
 	# This is temporary until a better container build method is developed
 	${GOENV} GOOS=linux go build ${GOBUILDFLAGS} -o ${BINFILE} ${MAINPACKAGE}
 
+# ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
+ENVTEST_K8S_VERSION = 1.23
+
+ENVTEST = $(shell pwd)/bin/setup-envtest
+.PHONY: envtest
+envtest: ## Download envtest-setup locally if necessary.
```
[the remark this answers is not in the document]
wanghaoran1988: @2uasimojo I did some updates:
1. rename envtest to setup-envtest, it's purpose is download/install the setup-envtest binary
2. Using --bin-dir to install the KUBEBUILDER_ASSETS to $(pwd)/bin

> Invoking setup-envtest here in this target rather than inline where it's consumed; setting a variable with the output; and using that variable where it's needed (on L260 in place of the $(shell ...) bit).

Not sure how to expose an variable in one make target, and consumed by another target, so I am not doing this atm.  we only need this when run 'make go-test', so I keeps it there.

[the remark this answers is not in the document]
2uasimojo: You can use `eval` to set a variable from within a target. Example:

```makefile
doit: bar
	echo $(FOO)

bar:
	$(eval FOO:=xyz)
```

```
[efried@efried tmp]$ make doit
echo xyz
xyz
```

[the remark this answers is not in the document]
wanghaoran1988: thanks, it works!

supreeth7:
> This looks solid to me. Have you tested it on both an old- and a new-osdk consuming repo yet?

I've been working on the migration side of things and have tested it with the most recent commit, works fine for upgraded osdk repo.

supreeth7:
openapi-gen is having some problems now.

If I use the make target, it generates a partially empty file.
```
$ make openapi-generate                                                                  
find gcp-project-operator/api -maxdepth 2 -mindepth 1 -type d | xargs -t -n1 -I% \
        openapi-gen --logtostderr=true \
                -i % \
                -o "" \
                -O zz_generated.openapi \
                -p % \
                -h /dev/null \
                -r "-"
xargs: warning: options --max-args and --replace/-I/-i are mutually exclusive, ignoring previous --max-args value
openapi-gen '--logtostderr=true' -i gcp-project-operator/api/v1alpha1 -o '' -O zz_generated.openapi -p gcp-project-operator/api/v1alpha1 -h /dev/null -r -
```
While if I run it explicitly, it works fine.
```
$ openapi-gen --logtostderr=true \                                                      
                  -i ./api/v1alpha1 \
                  -o "" \
                  -O zz_generated.openapi \
                  -p ./api/v1alpha1 \
                  -h /dev/null \
                  -r "-"
API rule violation: list_type_missing,./api/v1alpha1,ProjectClaimSpec,AvailabilityZones
API rule violation: list_type_missing,./api/v1alpha1,ProjectClaimStatus,Conditions
API rule violation: list_type_missing,./api/v1alpha1,ProjectReferenceStatus,Conditions
```

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -46,6 +46,9 @@ OPERATOR_IMAGE_URI=${IMG}
 OPERATOR_IMAGE_URI_LATEST=$(IMAGE_REGISTRY)/$(IMAGE_REPOSITORY)/$(IMAGE_NAME):latest
 OPERATOR_DOCKERFILE ?=build/Dockerfile
 REGISTRY_IMAGE=$(IMAGE_REGISTRY)/$(IMAGE_REPOSITORY)/$(IMAGE_NAME)-registry
+#The api dir that latest osdk generated
+NEW_API_DIR=$(PWD)/api
```
2uasimojo, line 50: I think `$(PWD)` is coming from your env, and may or may not be what you expect. This would probably work well as
```suggestion
NEW_API_DIR=./api
```

Similarly for the other places you're using `$(PWD)`.

If you actually do want to use the full path, consider evaluating it on the fly, like
```makefile
PWD := $(shell pwd)
```

2uasimojo (commented):
First, it looks like using `-I` to xargs gets you `-n1` for free, so you can omit the latter. But that's not what's causing the problem.

The `-t` flag to xargs is printing the command it's going to execute:

> openapi-gen '--logtostderr=true' -i gcp-project-operator/api/v1alpha1 -o '' -O zz_generated.openapi -p gcp-project-operator/api/v1alpha1 -h /dev/null -r -

Note that the parameter value being supplied to `-i` and `-p` is `gcp-project-operator/api/v1alpha1`. Whereas in the command you said was working:

> $ openapi-gen --logtostderr=true \                                                      
>                   -i ./api/v1alpha1 \
>                   -o "" \
>                   -O zz_generated.openapi \
>                   -p ./api/v1alpha1 \
>                   -h /dev/null \
>                   -r "-"

...it's `./api/v1alpha1`.

In the make target, that value is coming from `$(API_DIR)`. Assuming you're using this PR against the latest gcp-project-operator, that ought to be set te `$(NEW_API_DIR)`, which should be `$(PWD)/api`. AFAICT in make, `$(PWD)` isn't magic, so it's coming from your env. (What I can't figure out is how `find` is finding anything under that directory. Is your clone clean? `git clean -fxd`)

Try the suggestion inline and see if it helps.

boilerplate/openshift/golang-osd-operator/standard.mk
supreeth7: I did something similar earlier today, forgot to update it here sorry. That did the trick, works now. Thanks!

wanghaoran1988: thanks, ./api works.

wanghaoran1988:
@2uasimojo updated, could you take another look?

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -153,11 +161,34 @@ go-generate:
...
+endif
+
+.PHONY: controller-gen
+controller-gen: ## Download controller-gen locally if necessary.
+	$(call go-get-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@$(CONTROLLER_GEN_VERSION))
+
 .PHONY: op-generate
-op-generate:
-	# The artist formerly known as `operator-sdk generate crds`:
+op-generate: controller-gen
```
2uasimojo, line 189: Sorry, I just clued into this.

Whereas this will make it easier to run locally, it conflicts with the way we're [version-controlling controller-gen from the bp backing image](https://github.com/openshift/boilerplate/blob/9030cbfacb46509b1e38eac089d3428b282fb166/config/build_image-v2.0.0.sh#L41-L45). Basically, someone using the golang-osd-operator convention will be ignoring that config and using the version you're setting up here, whereas someone using a different convention would get what they expect from that image.

What happens if you use controller-gen 0.8.0 on a pre-osdkv1 repo? If it works, and doesn't result in any *functional* differences*, then I would suggest updating there (so that everyone is using 0.8.0) and removing this whole section.

If it doesn't work -- if you truly need 0.3.0 for old osdk and 0.8.0 for new -- then I suggest updating the backing image to
- install _both_ binaries with version suffixes
- symlink 0.3.0 to the unadorned `controller-gen` name as the default for backward compatibility
- twiddle this makefile to assume 0.8.0 exists at the appropriate path and either re-symlink it or set a var that points to the right version-suffixed binary

Either way you're going to need a boilerplate backing image update, which I suggest you do in a separate PR as it involves some fiddly tag manipulation. See #204 for a recent example.

*it's okay if there are trivial deltas that can easily be committed to the consuming repo without affecting form or function

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -153,11 +161,34 @@ go-generate:
...
 ifeq ($(CRD_VERSION), v1beta1)
-	cd pkg/apis; controller-gen crd paths=./... output:dir=../../deploy/crds
+	cd $(API_DIR); $(CONTROLLER_GEN) crd paths="./..." output:dir=$(PWD)/deploy/crds
```
2uasimojo, line 191: It looks like L191 is the same as L205, except for some quotes. Do those really matter? If not, suggest putting it outsid of the conditional and deduplicating.

wanghaoran1988: Done

boilerplate/openshift/golang-osd-operator/standard.mk
wanghaoran1988: Hi @2uasimojo we do need 0.3.0 for old, and 0.8.0 for new, cause it generate differences. 
PR updated, and I created https://github.com/openshift/boilerplate/pull/217

2uasimojo:
Thanks @wanghaoran1988, this looks great now. I approved #217 -- we can land this once it merges. Though it looks like it needs to be rebased first anyway.

supreeth7:
@2uasimojo Haoran's on a week long PTO. This is currently a blocker for something we've been working on so I'll be taking over from here.
I've rebased, I wonder if I did it right though? I haven't force pushed yet to revert just in case.
If needed, I'll be happy to close this one and open up a new PR with the same changes.

supreeth7:
Tested it with the latest o-sdk, all targets work fine.
Partially tested it with the old o-sdk, so far nothing breaking.
Going to unhold this, if any errors are noticed (likely none) will fix it in another series of PR(s).

/unhold

PR #212, opened 2022-04-27
make target for fips on osd operators

sedroche:
boilerplate support FIPS build

Jira: https://issues.redhat.com/browse/OSD-10744

- Adding a new flag to pass a fips build flag to operator builds
- new make target to create a standardised go file to import and enable fips when that build flag is used and ensure dockerfiles have the new flag set

Building on https://github.com/openshift/boilerplate/pull/204 which added the fips go toolset

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/README.md
+++ b/boilerplate/openshift/golang-osd-operator/README.md
@@ -104,3 +104,22 @@ Checks consist of:
 * `openapi-gen`. This is a no-op if your operator has no APIs.
 * `go generate`. This is a no-op if you have no `//go:generate`
   directives in your code.
+
+## FIPS (Federal Information Processing Standards)
+
+To enable FIPS in your build there is a `make ensure-fips` target.
+
+Add `FIPS_ENABLED=true` to your repos Makefile. Please ensure that this variable is added **before** including boilerplate Makefiles.
+
+e.g.
+```.mk
+FIPS_ENABLED=true
+
+include boilerplate/generated-includes.mk
+```
+
+`ensure-fips` will add a [fips.go](./fips.go) file in the same directory as the `main.go` file. (Please commit this file as normal)
+
+`fips.go` will import the necessary packages to restrict all TLS configuration to FIPS-approved settings.
+
+With `FIPS_ENABLED=true`, `ensure-fips` is always run before `make go-build`
--- a/boilerplate/openshift/golang-osd-operator/configure-fips.sh
+++ b/boilerplate/openshift/golang-osd-operator/configure-fips.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+
+set -e
+
+REPO_ROOT=$(git rev-parse --show-toplevel)
+CONVENTION_DIR="$REPO_ROOT/boilerplate/openshift/golang-osd-operator"
+PRE_V1_SDK_MANAGER_DIR="$REPO_ROOT/cmd/manager"
+
+if [[ -d "$PRE_V1_SDK_MANAGER_DIR" ]]
+then
+  MAIN_DIR=$PRE_V1_SDK_MANAGER_DIR
+else
+  MAIN_DIR=$REPO_ROOT
+fi
+
+echo "Writing fips file at $MAIN_DIR/fips.go"
+
+cp $CONVENTION_DIR/fips.go "$MAIN_DIR/fips.go"
--- a/boilerplate/openshift/golang-osd-operator/standard.mk
+++ b/boilerplate/openshift/golang-osd-operator/standard.mk
@@ -75,7 +75,13 @@ GOARCH?=$(shell go env GOARCH)
 # Consumers may override GOFLAGS_MOD e.g. to use `-mod=vendor`
 unexport GOFLAGS
 GOFLAGS_MOD ?=
-GOENV=GOOS=${GOOS} GOARCH=${GOARCH} CGO_ENABLED=0 GOFLAGS=${GOFLAGS_MOD}
+
+ifeq (${FIPS_ENABLED}, true)
+GOFLAGS_MOD+=-tags=fips_enabled
+GOFLAGS_MOD:=$(strip ${GOFLAGS_MOD})
+endif
+
+GOENV=GOOS=${GOOS} GOARCH=${GOARCH} CGO_ENABLED=0 GOFLAGS="${GOFLAGS_MOD}"
 
 GOBUILDFLAGS=-gcflags="all=-trimpath=${GOPATH}" -asmflags="all=-trimpath=${GOPATH}"
 
@@ -191,6 +197,10 @@ openapi-generate:
 .PHONY: generate
 generate: op-generate go-generate openapi-generate
 
+ifeq (${FIPS_ENABLED}, true)
+go-build: ensure-fips
+endif
+
 .PHONY: go-build
 go-build: ## Build binary
 	# Force GOOS=linux as we may want to build containers in other *nix-like systems (ie darwin).
@@ -272,3 +282,7 @@ opm-build-push: docker-push
 	OPERATOR_IMAGE_TAG="${OPERATOR_IMAGE_TAG}" \
 	OLM_CHANNEL="${OLM_CHANNEL}" \
 	${CONVENTION_DIR}/build-opm-catalog.sh
+
+.PHONY: ensure-fips
+ensure-fips:
+	${CONVENTION_DIR}/configure-fips.sh
--- a/boilerplate/openshift/golang-osd-operator/fips.go
+++ b/boilerplate/openshift/golang-osd-operator/fips.go
@@ -1,0 +1,15 @@
+// +build fips_enabled
+
+// BOILERPLATE GENERATED -- DO NOT EDIT
+// Run 'make ensure-fips' to regenerate
+
+package main
+
+import (
+	_ "crypto/tls/fipsonly"
+	"fmt"
+)
+
+func init() {
+	fmt.Println("***** Starting with FIPS crypto enabled *****")
+}


sedroche:
cc @dofinn

boilerplate/openshift/golang-osd-operator/configure-fips.sh
```diff
@@ -0,0 +1,32 @@
...
+
+func init() {
+  fmt.Println("***** Starting with FIPS crypto enabled *****")
+}
+EOF
+
+if ! grep -Fxq "ENV FIPS_ENABLED=true" build/Dockerfile
+then
+  echo "Writing to build/Dockerfile's to include FIPS_ENABLED=true"
+  sed -i '/^FROM quay.io\/app-sre\/boilerplate.*/a \\nENV FIPS_ENABLED=true' build/Dockerfile
```
sedroche, line 31: It puts `ENV FIPS_ENABLED=true` after the boilerplate builder `FROM` with a new line

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -272,3 +279,7 @@ opm-build-push: docker-push
 	OPERATOR_IMAGE_TAG="${OPERATOR_IMAGE_TAG}" \
 	OLM_CHANNEL="${OLM_CHANNEL}" \
 	${CONVENTION_DIR}/build-opm-catalog.sh
+
+.PHONY: enable-fips
+enable-fips:
```
2uasimojo, line 284: Is the plan to have the consuming repo's Makefile invoke this somehow? Thinking about it, I wonder if it makes more sense to set it up so the consuming repo can just set a variable -- e.g. `FIPS_ENABLED:=true` -- and the rest gets done under the covers automatically. E.g. I would envision this target (which I would rename `ensure-fips`, and make idempotent as commented above) being included as a dep automatically in the appropriate places, something like:

```makefile
ifeq (${FIPS_ENABLED}, true)
go-generate: ensure-fips
endif
```

...and similar for other existing targets. (Note that dep lines like this are additive. I.e.
```makefile
foo: bar
foo: baz
```
is equivalent to
```makefile
foo: bar baz
```
so the above works to conditionally add the dep to the already-defined target after the fact.)

boilerplate/openshift/golang-osd-operator/configure-fips.sh
sedroche: We have `if ! grep -Fxq "ENV FIPS_ENABLED=true" build/Dockerfile` on line 20 to check if it exist already

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -75,7 +75,14 @@ GOARCH?=$(shell go env GOARCH)
 # Consumers may override GOFLAGS_MOD e.g. to use `-mod=vendor`
 unexport GOFLAGS
 GOFLAGS_MOD ?=
-GOENV=GOOS=${GOOS} GOARCH=${GOARCH} CGO_ENABLED=0 GOFLAGS=${GOFLAGS_MOD}
+
+FIPS_ENABLED ?=
+ifeq (${FIPS_ENABLED}, true)
+GOFLAGS_MOD+=-tags=fips_enabled
```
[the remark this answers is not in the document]
sedroche: I did not know that, ty.
Do you want me to change this, we will be using an env var for  `FIPS_ENABLED`

boilerplate/openshift/golang-osd-operator/standard.mk
sedroche: The plan is to have each osd operator to have the option to opt into FIPS.
So once the operator updates to this version of boileplate, if they need FIPS they can run 'make enable-fips`  and they get these standardised files/flags etc.

FIPS will only be used in the dockerfile build and not locally

boilerplate/openshift/golang-osd-operator/configure-fips.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
+set -e
+
+REPO_ROOT=$(git rev-parse --show-toplevel)
+MANAGER_DIR=${REPO_ROOT}/cmd/manager
```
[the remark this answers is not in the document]
sedroche: Cheers, added a check for this

boilerplate/openshift/golang-osd-operator/configure-fips.sh
2uasimojo: D'oh, missed that, cool.

boilerplate/openshift/golang-osd-operator/standard.mk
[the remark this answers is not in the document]
2uasimojo: No changes necessary, jfyi.

boilerplate/openshift/golang-osd-operator/configure-fips.sh
```diff
@@ -0,0 +1,24 @@
...
+if [[ -d "$PRE_V1_SDK_MANAGER_DIR" ]]
+then
+  MAIN_DIR=$PRE_V1_SDK_MANAGER_DIR
+else
+  MAIN_DIR=$REPO_ROOT
+fi
+
+echo "Writing fips file at $MAIN_DIR/fips.go"
+
+cp $CONVENTION_DIR/fips.go "$MAIN_DIR/fips.go"
```
2uasimojo, line 18: I think you forgot to `git add` fips.go

boilerplate/openshift/golang-osd-operator/standard.mk
sedroche: That's pretty cool!

In my scenario here I would be adding a dep to go-build:
```.mk
ifeq (${FIPS_ENABLED}, true)
go-build: enable-fips
endif
```

The problem with having that variable set in the Makefile though is that on first run of `make go-build` locally the build will fail unless the user has the special go-toolset from the boilerplate image set as the GOROOT .
```.bash
$ make go-build
boilerplate/openshift/golang-osd-operator/configure-fips.sh
Writing fips file at /home/go/src/github.com/openshift/managed-upgrade-operator/cmd/manager/fips.go
# Force GOOS=linux as we may want to build containers in other *nix-like systems (ie darwin).
# This is temporary until a better container build method is developed
GOOS=linux GOARCH=amd64 CGO_ENABLED=0 GOFLAGS="-tags=fips_enabled" GOOS=linux go build -gcflags="all=-trimpath=/home/johnroche/go" -asmflags="all=-trimpath=/home/johnroche/go" -o build/_output/bin/managed-upgrade-operator ./cmd/manager
cmd/manager/fips.go:9:2: package crypto/tls/fipsonly is not in GOROOT (/home/sdk/go1.16.7/src/crypto/tls/fipsonly)
```

I was trying to avoid having the need for that go-toolset to build locally.

The current way the consuming repo just:
- Updates boilerplate (+ commit)
- run `make enable-fips` (+ commit)

// Local builds use non fips go-toolset
// Dockerbuilds use fips

Does that make sense? That's what I was thinking about with that, open to suggestions on smarter ways to do that.

boilerplate/openshift/golang-osd-operator/configure-fips.sh
sedroche: Yeah, that sounds like something I would do :facepalm: 
I'll commit it now.

boilerplate/openshift/golang-osd-operator/standard.mk
2uasimojo: You have described a canonical use case for [container-make](https://github.com/openshift/boilerplate/tree/master/boilerplate/openshift/golang-osd-operator#local-testing). It was designed exactly so that "local" builds could look just like CI and appsre builds, avoiding most of the problems that can arise from running on different platforms, with different (versions of) supporting tools installed, etc.

Put another way: you really shouldn't be counting on `make $target` for dev/test at all, not just because of FIPS.

Other than needing to install extra tools, is there some other reason local builds can't/shouldn't enable FIPS? E.g. does it significantly slow down the build? Does it refuse to build due to package entitlements?

sedroche: Let me double check with some more FIPS knowledgeable people. (MUO on ARO already uses FIPS) so
@hawkowl and @cblecker too

Any issues with/can we have the special go-toolset locally for dev?
Does running FIPS locally cause any issues?

sedroche: @2uasimojo I do agree with your approach though so i'll update with your suggestion.
qq on this, testing with the Makefile var, I had to put the var before the boilerplate includes.
e.g. In the consuming repo Makefile:
```.mk
FIPS_ENABLED=true

include boilerplate/generated-includes.mk
```
Is that acceptable or is there a way to do this I am unaware of?

2uasimojo: Fine by me 😁

sedroche: Perfect. And have updated with the suggested changes.
Let me know if you want the commit squashed.

2uasimojo (commented):
Cool.

I would just suggest updating the convention's [README](https://github.com/openshift/boilerplate/blob/master/boilerplate/openshift/golang-osd-operator/README.md) to indicate that this is a thing you can do, and how to do it, calling out in particular that you have to put `FIPS_ENABLED=true` **above** the `include ...generated...`.

2uasimojo:
Also, yes, please squash the commits since you're repushing. (Personally I like "one thing, one commit" in the git history when all is merged in, and this is one thing.)

sedroche:
@2uasimojo  README updated and commits squashed :+1:

boilerplate/openshift/golang-osd-operator/README.md
```diff
@@ -104,3 +104,22 @@ Checks consist of:
 * `openapi-gen`. This is a no-op if your operator has no APIs.
 * `go generate`. This is a no-op if you have no `//go:generate`
   directives in your code.
+
+## FIPS (Federal Information Processing Standards)
```
2uasimojo, line 108: :heart:

2uasimojo (commented):
/lgtm

Thanks!

PR #220, opened 2022-05-11
Setup KUBEBUILDER_ASSETS in go-test target

wanghaoran1988:
I have to move this into the 'go-test' target, or it will fail the first time you run it.
```
 make go-test            
bash: line 1: /home/haowang/go/src/github.com/openshift/gcp-project-operator/bin/setup-envtest: No such file or directory
go: creating new go.mod: module tmp
Downloading sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
Installed in /home/haowang/go/src/github.com/openshift/gcp-project-operator/bin/setup-envtest
KUBEBUILDER_ASSETS="" go test  github.com/openshift/gcp-project-operator github.com/openshift/gcp-project-operator/api/v1alpha1 github.com/openshift/gcp-project-operator/config github.com/openshift/gcp-project-operator/controllers github.com/openshift/gcp-project-operator/pkg/condition github.com/openshift/gcp-project-operator/pkg/configmap github.com/openshift/gcp-project-operator/pkg/gcpclient github.com/openshift/gcp-project-operator/pkg/util github.com/openshift/gcp-project-operator/pkg/util/errors github.com/openshift/gcp-project-operator/pkg/util/mocks github.com/openshift/gcp-project-operator/pkg/util/mocks/condition github.com/openshift/gcp-project-operator/pkg/util/mocks/controllers github.com/openshift/gcp-project-operator/pkg/util/mocks/gcpclient github.com/openshift/gcp-project-operator/pkg/util/mocks/projectclaim github.com/openshift/gcp-project-operator/pkg/util/mocks/structs

```

You can see KUBEBUILDER_ASSETS=""  is set to empty which is the reason why it fail the test for operator using latest osdk.

PR #221, opened 2022-05-11
Include openapi-gen and setup-envtest in the backing image

supreeth7:
1. The validation check on the PR [openshift/gcp-project-operator#177](https://github.com/openshift/gcp-project-operator/pull/177) is currently failing. This is due to the latest version of openapi-gen in which `github.com/go-openapi/spec` is replaced by `k8s.io/kube-openapi/pkg/validation/spec`. While `make validate` succeeds locally, the same fails when run containerized. 
Hence including openapi-gen v0.19.4 & v0.23.0 in the backing image to make it compatible with both the old & new operator-sdk versions.

2. Looking at setup-envtest
```
.PHONY: setup-envtest
setup-envtest: ## Download setup-envtest locally if necessary.
	$(call go-get-tool,$(SETUP_ENVTEST),sigs.k8s.io/controller-runtime/tools/setup-envtest@latest)
	$(eval KUBEBUILDER_ASSETS := "$(shell $(SETUP_ENVTEST) use $(ENVTEST_K8S_VERSION) -p path --bin-dir $(PWD)/bin)")

.PHONY: go-test
go-test: setup-envtest
	KUBEBUILDER_ASSETS=$(KUBEBUILDER_ASSETS) go test $(TESTOPTS) $(TESTTARGETS)

```
When run `make go-test`, this fails:

 ```
 $ make go-test
 KUBEBUILDER_ASSETS="" go test  github.com/openshift/gcp-project-operator ...
 ```
This is because shell bit is getting evaluated at "compile time", not at "runtime". To mitigate this I've included setup-envtest as part of the backing image and it is installed along with the other tools.

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/standard.mk
+++ b/boilerplate/openshift/golang-osd-operator/standard.mk
@@ -181,11 +181,19 @@ rm -rf $$TMP_DIR ;\
 }
 endef
 
+# Deciding on the binary versions
 CONTROLLER_GEN_VERSION = v0.8.0
 CONTROLLER_GEN = controller-gen-$(CONTROLLER_GEN_VERSION)
+
+OPENAPI_GEN_VERSION = v0.19.4
+OPENAPI_GEN = openapi-gen-$(OPENAPI_GEN_VERSION)
+
 ifeq ($(USE_OLD_SDK), TRUE)
-#If we are using the old osdk, we use the default controller-gen and it's v0.3.0 for now.
+#If we are using the old osdk, we use the default controller-gen and openapi-gen versions.
+# Default version is 0.8.0 for now.
 CONTROLLER_GEN = controller-gen
+# Default version is 0.19.4 for now.
+OPENAPI_GEN = openapi-gen
 endif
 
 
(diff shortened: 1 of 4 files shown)


supreeth7:
There's also `make coverage` check failing with the error `panic: mkdir /.cache: permission denied` which didn't show up at all while testing locally :wrench:

2uasimojo:
This looks fine as far as it goes, but since you're updating the backing image, you'll need to do the tag dance again to make pr-check pass.

The `/.cache` error is odd, but let's see if it persists after this is fixed.

supreeth7:
Thank you for your comments.
I'll wait for #220 to get merged first so that I can rebase and proceed with the tag dance.

supreeth7:
@2uasimojo following up on your comment (suggestion?) on #220, does this look reasonable?

Note: I haven't pushed the tags yet just to be on the safe side. The pr-check if expected to fail.

2uasimojo:
Yup, this looks good, except I would like to see the commit and PR messages updated to reflect what we're doing with setup-envtest as well as openapi-gen.

2uasimojo:
Can try close/reopen. Or I can just merge it by hand.

supreeth7:
Trying close/reopen.

2uasimojo:
Okay, merging manually...

PR #225, opened 2022-05-17
Generate CRD v1 by default

wanghaoran1988:
Most existing operators are using CRD version v1.

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/standard.mk
+++ b/boilerplate/openshift/golang-osd-operator/standard.mk
@@ -90,6 +90,7 @@ GOFLAGS_MOD ?=
 ifeq ($(HOME),/)
 export HOME=/tmp/home
 endif
+PWD=$(shell pwd)
 
 ifeq (${FIPS_ENABLED}, true)
 GOFLAGS_MOD+=-tags=fips_enabled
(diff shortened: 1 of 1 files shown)


supreeth7:
> > Most existing operators are using CRD version v1.
> 
> I would even go so far as to say they should _all_ be using CRDv1, as v1beta1 was [removed as of k8s v1.22](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#customresourcedefinition-v122). I took a quick look at a couple of OSD clusters (hives02 and hivep02) and they're at 1.23. Assuming the others are as well. Are there any customer clusters running older k8s?

I would rightly agree so, just checked and _all_ operators seem to be using CRD v1.

2uasimojo:
> _all_ operators seem to be using CRD v1.

Agree, based on a gh search for `CRD_VERSION` in `Makefile`s.

So can we refactor this PR to simply remove the v1beta1 handling entirely?

supreeth7:
@wanghaoran1988's away for the week again.
I can make the changes, let me do a quick test with MUO.

supreeth7:
@2uasimojo sorry that took a while, everything checks out, does this look good enough?

2uasimojo:
You should just be able to get rid of the test/case/convention/openshift/golang-osd-operator/04-op-generate-crd-fixup test case at this point.

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -203,24 +204,10 @@ CONTROLLER_GEN = controller-gen
 OPENAPI_GEN = openapi-gen
 endif
 
-
 .PHONY: op-generate
+## CRDv1beta1 is now deprecated
```
2uasimojo, line 208: IMO we don't need this comment. But if you must leave it, I would say "is no longer supported". It was deprecated for quite a while :)

PR #228, opened 2022-05-27
Use pwd as the repo dir

wanghaoran1988:
router-monitor-operator is running a openshift binary build to run it's integration tests, when you run "oc start-build --from-dir", the source code was copied into the pod without any git staff, so "git rev-parse" will fail, as we already have "PWD" variable defined in the standard Makefile, we can use it directly. 

See the error logs: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_route-monitor-operator/171/pull-ci-openshift-route-monitor-operator-master-test-integration/1529983195339034624/build-log.txt

```
go: downloading github.com/hashicorp/golang-lru v0.5.4
boilerplate/openshift/golang-osd-operator/configure-fips.sh
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
```

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/configure-fips.sh
+++ b/boilerplate/openshift/golang-osd-operator/configure-fips.sh
@@ -2,7 +2,7 @@
 
 set -e
 
-REPO_ROOT=$(git rev-parse --show-toplevel)
+REPO_ROOT=${PWD}
 CONVENTION_DIR="$REPO_ROOT/boilerplate/openshift/golang-osd-operator"
 PRE_V1_SDK_MANAGER_DIR="$REPO_ROOT/cmd/manager"
 


2uasimojo:
This isn’t the only thing that relies on `git rev-parse` or similar. Can RMO’s test use `git clone` instead of a raw copy? (Can you point me to the test in question?)

wanghaoran1988:
@2uasimojo It's creating a binary build and upload the source code to do a binary build, so there is no git repo staff https://github.com/openshift/route-monitor-operator/blob/master/hack/test-integration.sh#L58

2uasimojo:
I _guess_? Problem being that you can run `git rev-parse --show-toplevel` from any directory in the project, whereas `${PWD}` will produce different (and wrong) results in that scenario.

However, the consumer's `verify` check ought to catch it if the result of that were committed.

However, that's kind of brittle, no?

Also: the test is failing on this particular `git` command, but what's to say it won't fail on the _next_ one if this one gets resolved?

...So I poked around a bit and found (what I think is) a better solution. I'll post it in https://github.com/openshift/route-monitor-operator/pull/171.

2uasimojo:
Cool, https://github.com/openshift/route-monitor-operator/pull/171/commits/0a37ea0d5c692c7f3426f869946f7e88db82cb5b fixed the problem; we can close this now, right?

PR #229, opened 2022-05-31
Automate migration to an extent

supreeth7:
Currently migrating an OSD operator requires moving around a lot of files as per the [SOP](https://github.com/openshift/ops-sop/blob/master/operator_sdk/migrate_operator_sdk.md#making-it-boilerplate-compatible).
This PR is an attempt to automate some parts of it making it easier for others looking to migrate an operator.

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/standard.mk
+++ b/boilerplate/openshift/golang-osd-operator/standard.mk
@@ -79,6 +79,7 @@ endif
 
 GOOS?=$(shell go env GOOS)
 GOARCH?=$(shell go env GOARCH)
+GOBIN?=$(shell go env GOBIN)
 
 # Consumers may override GOFLAGS_MOD e.g. to use `-mod=vendor`
 unexport GOFLAGS
@@ -332,3 +333,18 @@ opm-build-push: docker-push
 .PHONY: ensure-fips
 ensure-fips:
 	${CONVENTION_DIR}/configure-fips.sh
+
+# You will need to export the forked/cloned operator repository directory as OLD_SDK_REPO_DIR to make this work.
+# Example: export OLD_SDK_REPO_DIR=~/Projects/My-Operator-Fork
+.PHONY: migrate-to-osdk1
+migrate-to-osdk1:
+ifndef OLD_SDK_REPO_DIR
+	$(error OLD_SDK_REPO_DIR is not set)
+endif
+	# Copying files & folders from old repository to current project
+	rm -rf config
+	rsync -a $(OLD_SDK_REPO_DIR)/deploy . --exclude=crds
+	rsync -a $(OLD_SDK_REPO_DIR)/pkg . --exclude={'apis','controller'}
+	rsync -a $(OLD_SDK_REPO_DIR)/Makefile .
+	rsync -a $(OLD_SDK_REPO_DIR)/.gitignore .
+	rsync -a $(OLD_SDK_REPO_DIR)/ . --exclude={'cmd','version','boilerplate','deploy','pkg'} --ignore-existing


boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -332,3 +333,27 @@ opm-build-push: docker-push
...
+
+.PHONY: migrate-tools
+migrate-tools:
+	# Installing controller-gen v0.8.0
+	go install sigs.k8s.io/controller-tools/cmd/controller-gen@${CONTROLLER_GEN_VERSION}
+	mv ${GOBIN}/controller-gen ${GOBIN}/controller-gen-${CONTROLLER_GEN_VERSION}
+
+	# Installing openapi-gen v0.23.0
+	go install k8s.io/code-generator/cmd/openapi-gen@${OPENAPI_GEN_VERSION}
+	mv ${GOBIN}/openapi-gen ${GOBIN}/controller-gen-${CONTROLLER_GEN_VERSION}
```
2uasimojo, line 345: For maximum portability, instead of installing these locally, I would recommend making the `migrate` target run `generate` through `container-make`. Suggestion below.

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -332,3 +333,27 @@ opm-build-push: docker-push
...
+
+.PHONY: migrate-files
+# You will need to export the forked/cloned operator repository directory as REPO to make this work.
+# Example: export REPO=~/Projects/My-Operator-Fork
```
2uasimojo, line 349: Suggest
- Adding a check to make sure this variable is defined. Otherwise `rsync` will end up trying to copy stuff from under `/`, which probably won't work, but could be pretty bad if it does. See L1-16 of this file for examples of such checks.
- Naming the variable something more descriptive, like `$(OLD_SDK_REPO_DIR)`.

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -332,3 +333,27 @@ opm-build-push: docker-push
...
+migrate-files:
+	rm -rf config
+	rsync -a $(REPO)/deploy .
+	rsync -a $(REPO)/config/config.go config/
+	rsync -a $(REPO)/config/templates/csv-template.yaml config/templates/
+	rsync -a $(REPO)/pkg/ ./pkg/ --exclude={'apis','controller'}
+	rsync -a $(REPO)/ . --exclude={'cmd','version'} --ignore-existing
+
+.PHONY: migrate
+migrate: migrate-files migrate-tools generate
```
2uasimojo, line 359: - Suggest naming this rule more descriptively, e.g. `migrate-to-osdk1`
- Per above, suggest running the `generate` through `container-make`:
```suggestion
migrate: migrate-files migrate-tools
	./boilerplate/_lib/container-make generate
```

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -332,3 +333,27 @@ opm-build-push: docker-push
...
+	mv ${GOBIN}/controller-gen ${GOBIN}/controller-gen-${CONTROLLER_GEN_VERSION}
+
+	# Installing openapi-gen v0.23.0
+	go install k8s.io/code-generator/cmd/openapi-gen@${OPENAPI_GEN_VERSION}
+	mv ${GOBIN}/openapi-gen ${GOBIN}/controller-gen-${CONTROLLER_GEN_VERSION}
+
+.PHONY: migrate-files
+# You will need to export the forked/cloned operator repository directory as REPO to make this work.
+# Example: export REPO=~/Projects/My-Operator-Fork
+migrate-files:
```
2uasimojo, line 350: Suggest naming this rule more descriptively, e.g. `migrate-osdk1-files`.

(Or if you take my other suggestions, you could just inline this stuff into the `migrate` rule :shrug:)

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -332,3 +333,27 @@ opm-build-push: docker-push
...
+	rm -rf config
+	rsync -a $(REPO)/deploy .
+	rsync -a $(REPO)/config/config.go config/
+	rsync -a $(REPO)/config/templates/csv-template.yaml config/templates/
+	rsync -a $(REPO)/pkg/ ./pkg/ --exclude={'apis','controller'}
+	rsync -a $(REPO)/ . --exclude={'cmd','version'} --ignore-existing
```
2uasimojo, line 356: Did you want to copy in boilerplate also?

boilerplate/openshift/golang-osd-operator/standard.mk
supreeth7: Inlining sounds like the better option.

boilerplate/openshift/golang-osd-operator/standard.mk
supreeth7: My bad, excluded boilerplate this time.

boilerplate/openshift/golang-osd-operator/standard.mk
supreeth7: Getting this weird error when trying to use `container-make`, any idea what might be causing this?

```
# Copying files & folders
rm -rf config
rsync -a /home/Supreeth/Work/Forks/pagerduty-operator/deploy . --exclude=crds 
rsync -a /home/Supreeth/Work/Forks/pagerduty-operator/config/config.go config/
rsync -a /home/Supreeth/Work/Forks/pagerduty-operator/config/templates/ config/templates/
rsync -a /home/Supreeth/Work/Forks/pagerduty-operator/pkg . --exclude={'apis','controller'}
rsync -a /home/Supreeth/Work/Forks/pagerduty-operator/ . --exclude={'cmd','version','boilerplate','deploy','pkg'} --ignore-existing
# Running make generate
./boilerplate/_lib/container-make generate

==============================
Starting the container
==============================
Error: stat .docker/config.json: no such file or directory
==ERROR== Couldn't start detached container
```

supreeth7: Also... I might be doing it wrong, but even though I've exported `OLD_SDK_REPO_DIR` it fails to parse in Makefile. Do I have to initialize it somewhere?

```
PHONY: migrate-to-osdk1
migrate-to-osdk1:
	ifndef OLD_SDK_REPO_DIR
	$(error OLD_SDK_REPO_DIR is not set)
	endif
	# Copying files & folders
	rm -rf config
	rsync -a $(OLD_SDK_REPO_DIR)/deploy . --exclude=crds 
	rsync -a $(OLD_SDK_REPO_DIR)/config/config.go config/
	...
```

```
make migrate-to-osdk1                                                                                                git:(master|✚8…18 
boilerplate/openshift/golang-osd-operator/standard.mk:342: *** OLD_SDK_REPO_DIR is not set.  Stop.
```

2uasimojo: Think of `ifndef`/`endif` as "precompiler directives". They're not interpreted as execution lines in a rule, but rather when the Makefile is parsed/loaded. I'm not actually sure _what_ `make` is making of your rule as written, but left-justifying those two lines fixes it:
```make
PHONY: migrate-to-osdk1
migrate-to-osdk1:
ifndef OLD_SDK_REPO_DIR
	$(error OLD_SDK_REPO_DIR is not set)
endif
	...
```
What this is actually doing is changing the _content_ of your rule at "compile-time" depending on whether the variable is set. If the variable is set, that `error` line is included (which of course means the rest of the rule isn't reached); if unset, it's excluded (so only the remainder of the rule is processed).

As for your `container-make` problem: do you get the same error when you run `container-make` manually from the shell?

supreeth7: Thanks for that detailed explanation, it works now. I grow a bit wiser every time I open a PR here.
As for the `container-make` problem, it runs all fine when run locally/manually from the shell.

On second thought, instead of automating generating manifests, I could ask the user to manually run `./boilerplate/_lib/container-make generate` in the SOP. 
This way the user can double-check if all files/folders have been copied over and the directory structure looks the way it is expected to (I've seen some operators having a different directory structure). Once sure, they can go ahead with manually running the `generate` command, this gives me a perception of having more control.

supreeth7:
Ah, `./boilerplate/_lib/container-make generate` is causing the test to fail:

```bash
boilerplate/openshift/golang-osd-operator/standard.mk:342: *** OLD_SDK_REPO_DIR is not set.  Stop.

==============================
The 'make' command failed! Starting a shell in the container for debugging. Just 'exit' when done.
```

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -332,3 +333,17 @@ opm-build-push: docker-push
 .PHONY: ensure-fips
 ensure-fips:
 	${CONVENTION_DIR}/configure-fips.sh
+
+# You will need to export the forked/cloned operator repository directory as OLD_SDK_REPO_DIR to make this work.
+# Example: export OLD_SDK_REPO_DIR=~/Projects/My-Operator-Fork
+.PHONY: migrate-to-osdk1
+migrate-to-osdk1:
+ifndef OLD_SDK_REPO_DIR
+$(error OLD_SDK_REPO_DIR is not set)
```
2uasimojo, line 342: Indent just this line.

As written, this will include the error line at the base level of the file rather than as part of the rule, so you would need to have the variable set always. (It's the same as what's happening on L1-16.)

Whereas if you indent this line, you're affecting how the *rule* gets "compiled".

So as
```suggestion
	$(error OLD_SDK_REPO_DIR is not set)
```
...you can think of it this way: With `OLD_SDK_REPO_DIR` *set* your rule becomes:
```make
.PHONY: migrate-to-osdk1
migrate-to-osdk1:
	$(error OLD_SDK_REPO_DIR is not set)
	# Copying files & folders
	rm -rf config
	rsync -a $(OLD_SDK_REPO_DIR)/deploy . --exclude=crds
	rsync -a $(OLD_SDK_REPO_DIR)/pkg . --exclude={'apis','controller'}
	rsync -a $(OLD_SDK_REPO_DIR)/.gitignore .
	rsync -a $(OLD_SDK_REPO_DIR)/ . --exclude={'cmd','version','boilerplate','deploy','pkg'} --ignore-existing
```
...and of course never gets past that first line *when the rule is invoked*.
Whereas with it *unset* the rule becomes:
```make
.PHONY: migrate-to-osdk1
migrate-to-osdk1:
	# Copying files & folders
	rm -rf config
	rsync -a $(OLD_SDK_REPO_DIR)/deploy . --exclude=crds
	rsync -a $(OLD_SDK_REPO_DIR)/pkg . --exclude={'apis','controller'}
	rsync -a $(OLD_SDK_REPO_DIR)/.gitignore .
	rsync -a $(OLD_SDK_REPO_DIR)/ . --exclude={'cmd','version','boilerplate','deploy','pkg'} --ignore-existing
```

boilerplate/openshift/golang-osd-operator/standard.mk
2uasimojo: If you're going that route, why not ask them to do a full bp [bootstrap](https://github.com/openshift/boilerplate/#quick-start) rather than copying the `boilerplate` directory in. The onboarding process does a little bit more than just lay down files under that directory.

boilerplate/openshift/golang-osd-operator/standard.mk
supreeth7: I knew something was off... (:facepalm:) thanks for clarifying.

boilerplate/openshift/golang-osd-operator/standard.mk
supreeth7: So with that change, the migration process would be something like:
1. Bootstrap boilerplate by subscribing to the golang-osd-operator convention by default.
2. Run `make migrate-to-osdk1` to copy over files.
3. Run `container-make generate` to generate manifests.

Does that look feasible?

2uasimojo: Yah, something like that. Could swap steps 1 and 2.

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -79,6 +79,7 @@ endif
 
 GOOS?=$(shell go env GOOS)
 GOARCH?=$(shell go env GOARCH)
+GOBIN?=$(shell go env GOBIN)
```
2uasimojo, line 82: This probably isn’t necessary anymore; but no harm.

2uasimojo (commented):
/lgtm

PR #232, opened 2022-07-13
Automatic creation of ":latest" tag for osd-container-image

abyrne55:
This PR adds automatic creation of the community-standard ":latest" tag to the `osd-container-image` convention. Essentially, this means that running `make osd-container-image-push` pushes the resulting container to two URIs: example.com/user_name/repo_name:v1234 (as before), and example.com/user_name/repo_name:latest. 

This PR partially addresses [OSD-11616](https://issues.redhat.com/browse/OSD-11616), where it was reported that a project using boilerplate (osd-network-verifier) no longer had a current/correct "latest" tag on its quay.io repo.

### DIFF
--- a/boilerplate/openshift/osd-container-image/standard.mk
+++ b/boilerplate/openshift/osd-container-image/standard.mk
@@ -38,6 +38,10 @@ IMAGE_VERSION := $(VERSION_MAJOR).$(VERSION_MINOR).$(COMMIT_NUMBER)-$(CURRENT_CO
 IMAGE=$(IMAGE_REGISTRY)/$(IMAGE_REPOSITORY)/$(IMAGE_NAME)
 IMAGE_TAG=v$(IMAGE_VERSION)
 IMAGE_URI?=$(IMAGE):$(IMAGE_TAG)
+# We generate IMAGE_URI_LATEST from IMAGE_URI just in case IMAGE_URI was explicitly set
+# by the user (e.g., as is done in app-sre-build-push.sh). Failing that, we fallback to
+# generating IMAGE_URI_LATEST by just appending ":latest" to IMAGE
+IMAGE_URI_LATEST?=$(or $(firstword $(subst :, ,$(IMAGE_URI))),$(value IMAGE)):latest
 DOCKERFILE ?=./build/Dockerfile
 
 
@@ -76,10 +80,12 @@ isclean:
 .PHONY: osd-container-image-build
 osd-container-image-build: isclean
 	${CONTAINER_ENGINE} build --pull -f $(DOCKERFILE) -t $(IMAGE_URI) .
+	${CONTAINER_ENGINE} tag $(IMAGE_URI) $(IMAGE_URI_LATEST)
 
 .PHONY: osd-container-image-push
 osd-container-image-push: osd-container-image-login osd-container-image-build
 	${CONTAINER_ENGINE} push ${IMAGE_URI}
+	${CONTAINER_ENGINE} push ${IMAGE_URI_LATEST}
 
 .PHONY: prow-config
 prow-config:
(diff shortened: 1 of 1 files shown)


boilerplate/openshift/osd-container-image/standard.mk
```diff
@@ -76,10 +80,12 @@ isclean:
 .PHONY: osd-container-image-build
 osd-container-image-build: isclean
 	${CONTAINER_ENGINE} build --pull -f $(DOCKERFILE) -t $(IMAGE_URI) .
+	${CONTAINER_ENGINE} tag $(IMAGE_URI) $(IMAGE_URI_LATEST)
```
boranx, line 83: we can add an additional -t flag to build process like the following:
```shell
${CONTAINER_ENGINE} build --pull -f $(DOCKERFILE) -t $(IMAGE_URI) -t $(IMAGE_URI_LATEST) .
```
so no need to call the bin second time.

boilerplate/openshift/osd-container-image/standard.mk
```diff
@@ -102,7 +108,9 @@ osd-container-image-build-push-one: isclean osd-container-image-login
 	@(if [[ -z "${IMAGE_URI}" ]]; then echo "Must specify IMAGE_URI"; exit 1; fi)
 	@(if [[ -z "${DOCKERFILE_PATH}" ]]; then echo "Must specify DOCKERFILE_PATH"; exit 1; fi)
 	${CONTAINER_ENGINE} build --pull -f $(DOCKERFILE_PATH) -t $(IMAGE_URI) .
+	${CONTAINER_ENGINE} tag $(IMAGE_URI) $(IMAGE_URI_LATEST)
```
boranx, line 111: similar as above

boranx (commented):
looks pretty straightforward and clean! thanks @abyrne55 
I've left an item that is optional, current form works well.

abyrne55:
/hold

Thanks for the heads up, @2uasimojo. I'll put a hold on this pending further discussion of whether or not we want :latest tags.

My two cents: even if we're not using them, :latest tags are a widely-agreed-upon standard. I'd hope that anyone deploying an image to prod would know not to use :latest, but a user casually evaluating a project on their local machine should be able to just `podman pull <repo>:latest` and trust that they got the latest version IMO. Right now, that's not the case for the repo that motivated this PR (i.e., quay.io/app-sre/osd-network-verifier), leading users of :latest to pull a 2-month-old release with 14 known vulnerabilities (although I don't mean to imply that this is boilerplate's fault, of course).

ritmun:
My 2c on the "latest" conversation -  Downstream pipelines shouldn't use _latest_ but devs may want to test the _latest_ image itself before releasing it. (They could also use the commit ID for it, but it's a matter of ease.)

It does seem misleading if we have old _latest_ tags on an image. So, we can either get rid of them from the image repo completely, or keep tagging latest per this change. Is the first one an option?

If not, per Eric's comment, 

> if you do decide to do this, it should be done in osd-golang-operator as well as osd-container-image,
 
Can this change also be added to the following for consistency? 

https://github.com/openshift/boilerplate/blob/2c2a7bf36f86e0bbd588f2854656de0130974f41/boilerplate/openshift/golang-osd-operator/standard.mk

2uasimojo:
> either get rid of them from the image repo completely, or keep tagging latest per this change. Is the first one an option?

I would definitely recommend that, yes. I know we did it to at least _some_ repos when we made that change, but would be unsurprised to learn we didn't get all of them. And of course, if a new project that previously used :latest were to cut over to boilerplate, that latest :latest would remain static from that point on.

I've got no skin in this decision, just wanted to make sure you had all the background before pulling the trigger :)

boilerplate/openshift/osd-container-image/standard.mk
abyrne55: Neat! Implemented in latest commit. Thanks for the feedback.

PR #234, opened 2022-07-21
OSD-11742, OSD-12367 - bug and CVE fixes for catalog operators

tonytheleg:
- updates `custom-catalog-build-push.sh` to check the build for any failures regarding non-existent replacements in the index, and if it failed, builds specifying `--from-index` flag targeting the latest catalog image (not all upstream operators specify this value)
- updates custom catalog make file to fix issues with using podman and it trying to pass `--config` into certain commands
- bumps opm version to v1.23.2 to resolve CVE's found using the older version of opm injected by the base container image it uses

### DIFF
--- a/boilerplate/openshift/custom-catalog-osd-operator/custom-catalog-build-push.sh
+++ b/boilerplate/openshift/custom-catalog-osd-operator/custom-catalog-build-push.sh
@@ -12,7 +12,7 @@ source $REPO_ROOT/boilerplate/_lib/common.sh
 
 function usage() {
     cat <<EOF
-    Usage: $0 REGISTRY_IMAGE_URI BASE_IMAGE_PATH
+    Usage: $0 REGISTRY_IMAGE_URI
     REGISTRY_IMAGE_URI is the complete image URI that needs to be built
 EOF
     exit -1
@@ -81,7 +81,6 @@ else
   do
     bundle_image=$( cat ${f} | jq -r .bundle_image )
     build_catalog_image ${bundle_image} ${REGISTRY_IMAGE_URI}
-    echo $?
     if [[ ${?} == 0 ]]; then
       echo "pushing image"
       cd ${REPO_ROOT}
--- a/boilerplate/openshift/custom-catalog-osd-operator/custom-catalog-build-push.sh
+++ b/boilerplate/openshift/custom-catalog-osd-operator/custom-catalog-build-push.sh
@@ -14,7 +14,6 @@ function usage() {
     cat <<EOF
     Usage: $0 REGISTRY_IMAGE_URI BASE_IMAGE_PATH
     REGISTRY_IMAGE_URI is the complete image URI that needs to be built
-    BASE_IMAGE_PATH is the base URI path in the container registry (ie. quay.io/app-sre/image)
 EOF
     exit -1
 }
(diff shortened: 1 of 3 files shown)


boilerplate/openshift/custom-catalog-osd-operator/custom-catalog-build-push.sh
```diff
@@ -30,14 +31,37 @@ function build_catalog_image() {
...
-    --container-tool ${CONTAINER_ENGINE_SHORT}
+  # some upstream bundles images are built specifying a "replaces" field, which means it builds on top of the 
+  # previous minor version. we need to check if the build fails for that reason and use another method
+  echo "building catalog image..."
+  BUILD=$(${opm_local_executable} index add \
+    --bundles ${bundle_image} \
+    --tag ${image_tag} \
+    --container-tool ${CONTAINER_ENGINE_SHORT} 2>&1)
+  
+  ERR_COUNT=$(echo $BUILD | grep -c 'non-existent replacement')
```
2uasimojo, line 42: I would think you *also* want to check the exit code of the command itself (`$?`). Surely it could fail for reasons other than "non-existent replacement"?

boilerplate/openshift/custom-catalog-osd-operator/custom-catalog-build-push.sh
```diff
@@ -30,14 +31,37 @@ function build_catalog_image() {
...
+      --container-tool ${CONTAINER_ENGINE_SHORT} \
+      --from-index "${BASE_IMAGE_PATH}:${CATALOG_LATEST_TAG}"
+  fi
 }
 
-[[ $# -eq 1 ]] || usage
+[[ $# -eq 2 ]] || usage
 
 REGISTRY_IMAGE_URI=$1
+BASE_IMAGE_PATH=$2
```
2uasimojo, line 64: Instead of needing to add a new argument, you *could* parse the tag off of `$REGISTRY_IMAGE_URI` instead:
```suggestion
BASE_IMAGE_PATH=${REGISTRY_IMAGE_URI%:*}
```

boilerplate/openshift/custom-catalog-osd-operator/standard.mk
```diff
@@ -10,7 +10,7 @@ REGISTRY_TOKEN?=$(QUAY_TOKEN)
 
 
 # Accommodate docker or podman
-CONTAINER_ENGINE=$(shell command -v podman 2>/dev/null || command -v docker 2>/dev/null)
+CONTAINER_ENGINE=$(shell command -v podman 2>/dev/null || echo docker --config=$(CONTAINER_ENGINE_CONFIG_DIR))
```
2uasimojo, line 13: nit: This works fine, but would be "safer" with some quotes:
```suggestion
CONTAINER_ENGINE=$(shell command -v podman 2>/dev/null || echo "docker --config=$(CONTAINER_ENGINE_CONFIG_DIR)")
```

2uasimojo (commented):
nothing show-stopping in here

boilerplate/openshift/custom-catalog-osd-operator/custom-catalog-build-push.sh
```diff
@@ -30,14 +31,37 @@ function build_catalog_image() {
...
+  echo "building catalog image..."
+  BUILD=$(${opm_local_executable} index add \
+    --bundles ${bundle_image} \
+    --tag ${image_tag} \
+    --container-tool ${CONTAINER_ENGINE_SHORT} 2>&1)
+  
+  ERR_COUNT=$(echo $BUILD | grep -c 'non-existent replacement')
+  
+  # Dump output of build since its stderr and won't show with the variable capture in place
+  echo -e "\n$BUILD\n"
```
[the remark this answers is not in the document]
tonytheleg: Interesting, I'll give that a shot! I was originally thinking something like `BUILD=$( command 2>&1 | tee /dev/tty )` but I wasn't sure if `/dev/tty` would work inside the Jenkins job or not

boilerplate/openshift/custom-catalog-osd-operator/custom-catalog-build-push.sh
tonytheleg: Yeah, good call. I was thinking I was lazily checking for any other errors based on the return of the function but since I'm running another command right after that wouldn't really work. Will add that. Thanks!

boilerplate/openshift/custom-catalog-osd-operator/custom-catalog-build-push.sh
tonytheleg: I was thinking of doing this originally and honestly didnt know which solution was better/cleaner so I just picked one. Fine with this method too, I didn't have a preference.

boilerplate/openshift/custom-catalog-osd-operator/custom-catalog-build-push.sh
2uasimojo: One of the solutions in the SO article noted elsewhere has a hack for getting the RC from the left side of the pipe.

boilerplate/openshift/custom-catalog-osd-operator/custom-catalog-build-push.sh
2uasimojo: Yeah, your call. Could also make the args base and tag, to avoid the duplication.

boilerplate/openshift/custom-catalog-osd-operator/custom-catalog-build-push.sh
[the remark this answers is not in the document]
karthikperu7: @2uasimojo that is handy :)

karthikperu7:
Thanks for reviewing @2uasimojo and thanks for adding this fix @tonytheleg . I am happy for this to go through when you are ready. Although I dont have merge rights on this repo, I wonder if we can have OWNERS file for each boilerplate type like we have for managed cluster config....

tonytheleg:
Thanks @2uasimojo for all the suggestions. I think I've got everything implemented and local tests are working with all of our current catalog operators. Would you mind reviewing? Thank you

boilerplate/openshift/custom-catalog-osd-operator/custom-catalog-build-push.sh
```diff
@@ -30,14 +30,48 @@ function build_catalog_image() {
...
+  echo "building catalog image..."
 
+  # preserve logging output from command by duplicating the file descriptor
+  exec 5>&1
+  BUILD=$(${opm_local_executable} index add \
+    --bundles ${bundle_image} \
+    --tag ${image_tag} \
+    --container-tool ${CONTAINER_ENGINE_SHORT} 2>&1 | tee /dev/fd/5; exit ${PIPESTATUS[0]})
+  RC=$?
+  ERR_COUNT=$(echo $BUILD | grep -c 'non-existent replacement')
```
2uasimojo, line 44: Just a note here that `ERR_COUNT` will likely only ever be 0 or 1 unless you quote `"$BUILD"`. The way you're using it, that will still work, but jfyi.

boilerplate/openshift/custom-catalog-osd-operator/custom-catalog-build-push.sh
```diff
@@ -30,14 +30,48 @@ function build_catalog_image() {
...
+  fi
 }
 
 [[ $# -eq 1 ]] || usage
 
 REGISTRY_IMAGE_URI=$1
+BASE_IMAGE_PATH=${REGISTRY_IMAGE_URI%:*}
 
 if image_exists_in_repo "${REGISTRY_IMAGE_URI}"; then
   echo "Custom catalog image for the latest operator version already exists in the registry"
```
2uasimojo, line 77: NB: Since you removed `set -e`, if `build_catalog_image` were to fail (return nonzero), the loop below would silently no-op for that iteration. As currently written, all failure paths in `build_catalog_image` will exit, so it's not an issue rn; jfyi for future-proofing.

2uasimojo (commented):
Cool! I'm taking it on faith that your opm commands are doing the right thing. Otherwise this looks fine to me.

/lgtm

PR #235, opened 2022-07-25
Osd 12367 fix build error check

tonytheleg:
- updates the string to check errors for as it has updated in the newer version of opm being used now (did not catch this as build tests were done before the code changes so i know the version of opm works...just bad error checking)
- made some of the other fixes as mentioned in the previous [PR](https://github.com/openshift/boilerplate/pull/234#discussion_r929004459)

### DIFF
--- a/boilerplate/openshift/custom-catalog-osd-operator/custom-catalog-build-push.sh
+++ b/boilerplate/openshift/custom-catalog-osd-operator/custom-catalog-build-push.sh
@@ -41,7 +41,7 @@ function build_catalog_image() {
     --tag ${image_tag} \
     --container-tool ${CONTAINER_ENGINE_SHORT} 2>&1 | tee /dev/fd/5; exit ${PIPESTATUS[0]})
   RC=$?
-  ERR_COUNT=$(echo $BUILD | grep -c 'non-existent replacement')
+  ERR_COUNT=$(echo $BUILD | grep -c 'replaces nonexistent bundle')
     
   if [[ ${RC} > 0 ]] && [[ ${ERR_COUNT} == 0 ]]; then
     echo "adding bundle failed"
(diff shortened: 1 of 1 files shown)


tonytheleg:
/assign @2uasimojo

Eric, mind taking a look at this one too? The error check was my mistake, I built the images before the code changes the other week, and didn't notice the error string changed. I made the other recommended changes you mentioned in last PR though so I figured I'd bug you on this one. Thanks!

boilerplate/openshift/custom-catalog-osd-operator/custom-catalog-build-push.sh
```diff
@@ -36,12 +36,12 @@ function build_catalog_image() {
 
   # preserve logging output from command by duplicating the file descriptor
   exec 5>&1
-  BUILD=$(${opm_local_executable} index add \
+  BUILD="$(${opm_local_executable} index add \
```
2uasimojo, line 39: so what I meant was quoting `"$BUILD"` when you're going to echo it to grep. Viz:

```
[efried@efried hive]$ x=`printf 'foo\nbar\nfoo\nbaz\n'`
[efried@efried hive]$ echo $x
foo bar foo baz
[efried@efried hive]$ echo $x | grep -c foo
1
[efried@efried hive]$ echo "$x"
foo
bar
foo
baz
[efried@efried hive]$ echo "$x" | grep -c foo
2
```

tonytheleg: Ahhh that makes more sense, I didn't see what you were talking about but assumed you knew what was up haha. Got this fixed. Thanks.

2uasimojo (commented):
/lgtm

PR #238, opened 2022-08-22
Ignoring mock clients for codecov

ravitri:
Based on https://docs.codecov.com/docs/ignoring-paths#ignoring-specific-file-names, adding ignore path for any files/folders under `mocks` directory. This would improve the coverage for some of the repositories which have mock clients generated using `mockgen` for container-runtime or any other interfaces.


2uasimojo:
Yah, this should work.

Do any of the projects generate `zz_generated_*` files? That might be more low-hanging fruit (for a subsequent PR).

/lgtm

> This would improve the coverage

Just to be super nitpicky, we're not improving code coverage. Just boosting the numbers by making the denominator smaller.

PR #240, opened 2022-08-29
Add tools target

supreeth7:
Currently, boilerplate consuming repositories are of different operator-sdk versions; v0.x or v1.x. 
This requires specific versions of generator binaries that the developer has to manually install and configure (rename) to be compatible with the boilerplate's standard make commands.
This PR adds a new target called 'tools' to aid in the local installation of binaries.
<hr>

Resolves: [OSD-13024](https://issues.redhat.com/browse/OSD-13024)

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/README.md
+++ b/boilerplate/openshift/golang-osd-operator/README.md
@@ -14,6 +14,7 @@ This convention is suitable for both cluster- and hive-deployed operators.
 The following components are included:
 
 ## `make` targets and functions.
+
 **Note:** Your repository's main `Makefile` needs to be edited to include the
 "nexus makefile include":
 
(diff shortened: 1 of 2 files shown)


2uasimojo (commented):
Hm. I get where this is going, but IMO we don't want encourage people to try to get builds working their "local" environment rather than via `container-make`. This is a slippery slope -- if we do this for controller-gen and openapi-gen, we should do it for kustomize and mockgen and go-bindata and yq and python libs and even go itself...

supreeth7:
This is a tentative solution to help with the migration work ([ref](https://issues.redhat.com/browse/SDE-1972)) that is ongoing, there have been concerns regarding local environment setup with the introduction of versioned generators.

I would ideally like to make changes to the build image and make the newer versions of generators as default once all operators have been migrated with backward compatibility for the older generators. Cutting a new release now would disrupt the flow for those working on various migration projects. 

Hence proposing a tentative solution to locally setup critical (generators + envtest) binaries.

2uasimojo:
Two main things to unpack there.
1. We've already set things up so the migration can be smooth. Projects using old osdk simply set `USE_OLD_SDK:=TRUE` in their Makefile. When the project is updated to use the new one, remove that line and everything Just Works™.
2. This:
> devs like to use `make test` & `make generate` than `container-make`, this has been called out. :)

Devs need to understand that, no matter how much spaghetti code we put in standard.mk, no matter how many platform/arch/os-specific rules and conditionals we mash in, running those commands locally is *not* guaranteed to give the same results as CI will. This was a significant issue before boilerplate, and one of the main reasons for its inception -- and specifically for the creation of `container-make`. If we start down this path, we will *always* be chasing that last 10% of compatibility.

What is the objection to `container-make`? Typing the extra chars to invoke it, or putting it in their `$PATH`? So they don't have to wait for the container image to download (only the first time!)? So they don't have to wait for the container to start?

supreeth7:
Easing the local environment setup isn't a viable solution as we'd be pushing people down the wrong path when ideally we'd want them to use `container-make` so that their local results are consistent with what the CI produces.

On that note, I'll be closing this PR.

PR #241, opened 2022-09-01
Make env var optional for operator containers

mjlshen:
Using this logic to make environment vars optional as they aren't strictly required for MNMO to function https://github.com/openshift/managed-node-metadata-operator/pull/39

```python
env = deploy['spec']['template']['spec']['containers'][0].get('env')
if env:
    # env var exists on first container
else:
    # env var doesn't exist on first container
```

The error we were getting was essentially (see MNMO PR for more context if needed):
```
11:32:21 Traceback (most recent call last):
11:32:21   File "./boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py", line 322, in <module>
11:32:21     env = deploy['spec']['template']['spec']['containers'][0]['env']
11:32:21 KeyError: 'env'
```


2uasimojo (commented):
Yah, seems reasonable

/lgtm

PR #242, opened 2022-09-06
Add container-make targets

supreeth7:
This PR adds `make container-{target}` targets to the convention Makefile making boilerplate's container-make more discoverable.

_The only con to this is duplication of targets, but given the enhanced discovery easing local testing, this should be plausible._


2uasimojo (commented):
Yah, this seems like a reasonable tradeoff.

/lgtm

PR #243, opened 2022-09-30
update ubi image

katherinelc321:
this updates the ubi image to address a couple of medium vulnerabilities 
https://issues.redhat.com/browse/OSD-13308
the image has been built locally and quay scan is green
@tonytheleg and i have tested in MUO and RBAC operator with no issue

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/csv-generate/catalog-build.sh
+++ b/boilerplate/openshift/golang-osd-operator/csv-generate/catalog-build.sh
@@ -67,7 +67,7 @@ FROM quay.io/openshift/origin-operator-registry:4.10.0 AS builder
 COPY $SAAS_OPERATOR_DIR manifests
 RUN initializer --permissive
 
-FROM registry.access.redhat.com/ubi8/ubi-micro:8.5-836
+FROM registry.access.redhat.com/ubi8/ubi-micro:latest
 
 COPY --from=builder /bin/registry-server /bin/registry-server
 COPY --from=builder /bin/grpc_health_probe /bin/grpc_health_probe
--- a/boilerplate/openshift/golang-osd-operator/csv-generate/catalog-build.sh
+++ b/boilerplate/openshift/golang-osd-operator/csv-generate/catalog-build.sh
@@ -67,7 +67,7 @@ FROM quay.io/openshift/origin-operator-registry:4.10.0 AS builder
 COPY $SAAS_OPERATOR_DIR manifests
 RUN initializer --permissive
 
-FROM registry.access.redhat.com/ubi8/ubi-micro:latest
+FROM registry.access.redhat.com/ubi8/ubi-micro:8.6-484
 
 COPY --from=builder /bin/registry-server /bin/registry-server
 COPY --from=builder /bin/grpc_health_probe /bin/grpc_health_probe


boilerplate/openshift/golang-osd-operator/csv-generate/catalog-build.sh
```diff
@@ -67,7 +67,7 @@ FROM quay.io/openshift/origin-operator-registry:4.10.0 AS builder
 COPY $SAAS_OPERATOR_DIR manifests
 RUN initializer --permissive
 
-FROM registry.access.redhat.com/ubi8/ubi-micro:8.5-836
+FROM registry.access.redhat.com/ubi8/ubi-micro:latest
```
2uasimojo, line 70: There are several reasons we've traditionally avoided using `latest` (or other "floating") tags -- not least of which is that it leaves you no easy way to tell what rev of the image you've built with after the fact. I'd let @jharrington22 have the final word here, but my recommendation would be to identify and use an explicit tag that has the fixes you want to pick up.

tonytheleg: We can substitute latest for a specific tag if desired for this PR, I've got one that's verified. The push to latest was to match what most of the operators are doing so that CVE's found can generally be fixed by creating an empty commit in the operator repo to trigger a build using the latest which is where the CVE's are coming in. Not as crucial to be latest now but there is an [epic](https://issues.redhat.com/browse/SDE-2251) to try and have automated builds to help reduce CVE's overall since FedRAMP is going to have strict timelines on when things must be remediated. Just let us know @jharrington22  if you want us to specify a version.

dustman9000: I'd also prefer using a specific tag (whatever latest is currently pointing to) as to make sure that future builds of boilerplate don't introduce breakages.

2uasimojo: > most of the operators are doing so that CVE's found can generally be fixed by creating an empty commit in the operator repo to trigger a build using the latest which is where the CVE's are coming in

I disagree with this as a strategy. It may seem easier on the surface, but:
- You're implicitly getting new content (maybe! sometimes!) when you push a commit for any reason. Ideally that's a no-op. Sometimes it picks up a fix that's "good" -- but since you didn't know you were doing it, it's hard to figure out _where_ you picked up that fix later on if you need to. And sometimes you pick up a change that breaks something -- hopefully in a way you can see immediately, but maybe not...
- If you have to create a commit anyway, why not create a commit that explicitly updates the tag?

IMHO this reasoning should supersede "do this because other OSD operators are doing it". (For that matter, there was a point where James and I were heavily pushing getting rid of floating tags across OSD for these reasons.)

tonytheleg: All fair points, my hesitance with latest was how many operators could be impacted by the change which by using a tag I guess would only be a worry on the first change to a tag. @katherinelc321 would you mind updating the PR to use `registry.access.redhat.com/ubi8/ubi-micro:8.6-484`

2uasimojo: It may also be worth considering putting that base image and its tag into boilerplate itself, so consuming repositories could pick up the most recent blessed image with a boilerplate update. (That's assuming all consumers share a common base image in this way -- though it can easily be made overrideable per repo as well.)

jharrington22: Thanks folks yes lets please be explicit on the container tag we want to pull. `latest` has many downsides, primarily you are not explicitly requesting any specific container version so as mentioned by @2uasimojo debugging and tracing is difficult through pipelines and on cluster and a added issue is caching. `latest` will be cached in the event it can't be pulled giving a false sense of security that we have the "latest" version.

I like @2uasimojo's idea of a UBI version reference in boilerplate, that could be in a separate PR though and maybe mentioned in https://issues.redhat.com/browse/SDE-2251 - I'm going to drop a comment on that RE `latest` tags anyway.

jharrington22 (approved):
/lgtm

PR #244, opened 2022-11-01
`opm-build-push` compatibility fixes

Ajpantuso:
### Summary

- Allows `opm-build-push` to work with an `image_builder` including a `--config` flag i.e. `docker --config ...`
- Fixes OLM incompatibilities caused by catalog images built with old OPM versions, see https://github.com/operator-framework/operator-registry/issues/619, in particular catalog images built with `v1.15.2` will fail on OCP 4.12+

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
+++ b/boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
@@ -71,7 +71,7 @@ function check_bundle_contents_cmd() {
 # Check we are running an opm supported container engine
 function check_opm_supported_container_engine() {
     local image_builder=${1}
-    if [[ "$image_builder" != "docker" && "$image_builder" != "podman" ]]; then
+    if [[ "$image_builder" != docker* && "$image_builder" != "podman" ]]; then
         # opm error messages are obscure. Let's make this clear
         log "image_builder $image_builder is not one of docker or podman"
         return 1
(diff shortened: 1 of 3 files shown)


2uasimojo (commented):
OPM version matches what's in the custom-catalog-osd-operator convention, so presumably it's cool ✓

/lgtm

boilerplate/openshift/golang-osd-operator/build-opm-catalog.sh
```diff
@@ -71,7 +71,7 @@ function check_bundle_contents_cmd() {
 # Check we are running an opm supported container engine
 function check_opm_supported_container_engine() {
     local image_builder=${1}
-    if [[ "$image_builder" != "docker" && "$image_builder" != "podman" ]]; then
+    if [[ "$image_builder" != docker* && "$image_builder" != "podman" ]]; then
```
2uasimojo, line 74: I wonder if it would be better to have a separate `$CONTAINER_ENGINE_FLAGS` var rather than glomming the args into `$CONTAINER_ENGINE`. But this is a reasonable quick fix.

Ajpantuso: Ideally the plan is to move DVO towards using `build-push` like other OSD operators. The issue is that currently https://github.com/app-sre/deployment-validation-operator distributes an OSD operator, an app-sre operator only deployed to app-sre clusters, and a community operator all of which overlap in some way in terms of build artifacts. Given the need to fix this before 4.12 a quick solution is preferred in the interim. Thanks.

PR #246, opened 2022-11-03
Release boilerplate v3.0.0 for Operator-SDK v1.25.0, Go 1.19

mjlshen:
Done as a part of [OSD-13790](https://issues.redhat.com/browse/OSD-13790)

Primarily to unblock [OSD-13437](https://issues.redhat.com//browse/OSD-13437) (but also this will allow for other operators to migrate to Operator-SDK v1.25.0 when they wish, which is the first version to support/expect Go 1.19):
* The Go 1.19 image is tied to OCP 4.12 (we tried 4.11, but it's not available)
* We copied `build_image-v2.0.0.sh` to `build_image-v3.0.0.sh`, feel free to diff, but in summary:
  * `golangci_lint` updated to "1.50.0" (for operator-sdk v1.25.0)
  * `gh` 1.10.3 to 2.19.0
  * `kustomize` updated to v4.5.5
  * `controller-gen` updated to have v0.10.0 (for operator-sdk v1.25.0)
  * `mockgen` updated to v1.6.0 (for operator-sdk v1.25.0)


config/build_image-v3.0.0.sh
```diff
@@ -0,0 +1,126 @@
+#!/usr/bin/env bash
+
+# This script comprises everything up to and including the boilerplate
+# backing image at version image-v2.0.0. It is used when performing a
```
2uasimojo, line 4:
```suggestion
# backing image at version image-v3.0.0. It is used when performing a
```

config/build_image-v3.0.0.sh
```diff
@@ -0,0 +1,126 @@
...
+# full build in the appsre pipeline, but is bypassed during presubmit CI
+# in prow to make testing faster there. As such, there is a (very small)
+# possibility of those behaving slightly differently.
```
2uasimojo, line 7: Please edit the corresponding comment in build_image-v2.0.0.sh to look like the one at the top of build_image-v1.0.0.sh (i.e. the bit about "unused, kept for posterity")

config/Dockerfile
```diff
@@ -1,5 +1,10 @@
...
+# TODO: temporary until image is built, should be: FROM registry.ci.openshift.org/openshift/boilerplate:image-v3.0.0
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.19-openshift-4.12
+
+# TODO: remove this COPY/RUN
+COPY build_image-v3.0.0.sh /build-temp.sh
+RUN /build-temp.sh
 
 COPY build.sh /build.sh
-RUN /build.sh && rm -f /build.sh
+RUN /build.sh
```
2uasimojo, line 10: I think we also want to consolidate the contents of build.sh into build_image-v3.0.0.sh and get rid of build.sh for this initial cut of v3.0.0.

Then, as you hint in your TODOs, cut a 3.0.1 with an empty build.sh that starts FROM 3.0.0. It'll be identical, but will get us back to where CI doesn't need to build the entire image from scratch, making runs faster.

mjlshen: Makes sense, I was wondering why it had been split out - done!

2uasimojo (commented):
Yah, this looks right. Will keep an eye out for passing pr-check once tags are pushed.

Also consider a doc change in a separate PR updating the process for pushing tags upstream :(

dustman9000:
I've pushed the `v3.0.0` to the repo, hopefully this will fix the PR checks. Thanks for the contributions!

2uasimojo:
Tags will need to be force-re-pushed for pr-check to proceed.

mjlshen:
Yeah the tag got force-re-pushed - maybe image-v3.0.0 does now exist and now I need a image-v3.0.1? A confusing knot to unwind on Monday haha

mjlshen:
On hold again until I can get this latest commit re-tagged

2uasimojo:
Does pr-check check the tag setup *before* running the actual test suite? That's frustrating, considering how awkward it is to update the tags. I would totally support a PR to swap the order of those things, so you could iterate on the tests and then re-tag when they go green.

mjlshen:
/retest
...maybe I did it too early

2uasimojo:
also sometimes CI caches results and you have to log in and delete builds to get a new run

mjlshen:
/retest

This is the one

mjlshen:
/retest

this is really the one

mjlshen:
/retest

the one retest to rule them all

test/case/convention/openshift/golang-osd-operator/01-generated-files-checker
```diff
@@ -2,45 +2,33 @@
...
-    err "Files have been generated while the project wasn't committed..."
-fi
-
-# Check failure when project is committed but generated content not generated/committed
-git add -A
-git commit -m "initial commit"
-
-make generate-check && err "make generate-check should have failed" || echo "Generated files are not committed. Checks failed as expected"
+# Generate files and check if the generated files matches the expected content
+make generate
```
2uasimojo, line 22: So... why can't we keep the `generate-check`s in here?

It's possible you'll need to update more of the files in src/test/ to match whatever the newer generators are putting out (like you did for file-generate).

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -191,20 +191,8 @@ rm -rf $$TMP_DIR ;\
...
-OPENAPI_GEN_VERSION = v0.23.0
-OPENAPI_GEN = openapi-gen-$(OPENAPI_GEN_VERSION)
-
-ifeq ($(USE_OLD_SDK), TRUE)
-#If we are using the old osdk, we use the default controller-gen and openapi-gen versions.
-# Default version is 0.3.0 for now.
 CONTROLLER_GEN = controller-gen
-# Default version is 0.19.4 for now.
 OPENAPI_GEN = openapi-gen
-endif
```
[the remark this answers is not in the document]
mjlshen: Yeah, to all of the old versions every operator was recently migrated to Operator-SDK v1.22 at a minimum in SDE-1972. The old versions are no longer needed - if there are any remaining consumers that need the old versions, the strategy is that they need to upgrade.

I can definitely call this out in the PR comments

2uasimojo:
Okay, given you've said we no longer need to cater to old operatorsdk, this is safe to land. You can edit the PR message any time, and fix the unit test in a followon (so we don't need to muck with tags here anymore!)

/lgtm

2uasimojo:
Oh, please also update (in a fup) the [Build Images docs](https://github.com/openshift/boilerplate/#build-images) to make things a bit easier for future-you. At the least, note that you have to collar a team lead to get upstream tags pushed.

test/case/convention/openshift/golang-osd-operator/01-generated-files-checker
mjlshen: It's a mix of me making file-generate a better example project and thinking that at some point the behavior changed and this test wasn't even functioning as intended, e.g.

1. `make generate-check` only should fail as expected because the original file-generate operator didn't have its openapi file generated, so then this would always fail. In the new file-generate operator, I have the correct openapi file already in place.
2. `make generate-check` already includes `make isclean`, so the second check is redundant
3. None of those files: `src/test/test-generation`, `srcs/test/subfolder/test-generation-subfolder`, nor `pkg/apis/mygroup/v1alpha1/zz_generated.deepcopy.go` would ever exist - this must be a check based on older logic
4. Since `make generate-check` in step 1 calls `make generate`, and then any generated files get committed, it didn't make sense why this final `make generate-check` should fail.

So I decided to just generate the files, keep the existing simple diffs/existence checks, and keep the useful `make generate-check` tests afterwards.

boilerplate/openshift/golang-osd-operator/standard.mk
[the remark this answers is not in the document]
2uasimojo: Oh, I forgot to mention: there are a few more instances of the `USE_OLD_SDK` var that I guess we can get rid of now.

test/case/convention/openshift/golang-osd-operator/01-generated-files-checker
2uasimojo: Yeah, looking at the rules and this test again, it didn't make a whole lot of sense. If you got it working, I'm happy :)

PR #247, opened 2022-11-11
Improving documentation around new image tags, cleanup old Operator-SDK logic

mjlshen:
[OSD-13437](https://issues.redhat.com//browse/OSD-13437): Continuing cleanup work as a result of #246 before releasing a v3.0.1 

* Tidying up the README a bit and adding a callout that we need a team lead to help push tags
* Cleaning up some logic around old operator-sdk versions in the `golang-osd-operator` convention
* Sharing the `ALLOW_DIRTY_CHECKOUT` workaround in the error message of `make isclean`
* Updating fips logic as now we just need to set an environment variable `GOEXPERIMENT=boringcrypto` https://github.com/golang/go/issues/51940#issuecomment-1206852980

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/standard.mk
+++ b/boilerplate/openshift/golang-osd-operator/standard.mk
@@ -46,9 +46,6 @@ OPERATOR_IMAGE_URI=${IMG}
 OPERATOR_IMAGE_URI_LATEST=$(IMAGE_REGISTRY)/$(IMAGE_REPOSITORY)/$(IMAGE_NAME):latest
 OPERATOR_DOCKERFILE ?=build/Dockerfile
 REGISTRY_IMAGE=$(IMAGE_REGISTRY)/$(IMAGE_REPOSITORY)/$(IMAGE_NAME)-registry
-#The api dir that latest osdk generated
-NEW_API_DIR=./api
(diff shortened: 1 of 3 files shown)


boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -71,11 +68,7 @@ REGISTRY_TOKEN ?=
 
 BINFILE=build/_output/bin/$(OPERATOR_NAME)
 MAINPACKAGE = ./
-API_DIR = $(NEW_API_DIR)
-ifeq ($(USE_OLD_SDK), TRUE)
-MAINPACKAGE = ./cmd/manager
-API_DIR = ./pkg/apis
-endif
+API_DIR = ./api
```
2uasimojo, line 71: Or just inline it. Is it used anywhere else?

2uasimojo (commented):
This looks great. One small suggestion inline

boilerplate/openshift/golang-osd-operator/standard.mk
2uasimojo: All three of these, actually.

Unless they're overridden in consuming projects?

mjlshen: They shouldn't be overridden since we should all have the same structure, so +1

2uasimojo (commented):
/lgtm

thanks!

PR #254, opened 2022-12-07
Add dependabot configuration for OSD Operators

dustman9000:
This adds dependabot configuration to perform weekly maintenance of the operator's base image.

Ref: https://issues.redhat.com/browse/SDE-2450


cblecker (commented):
This looks great!

A couple other considerations:
The Dockerfile in build/ would need to be updated with a tag in the format of `:8.6-484` instead of `:latest` for this to actually pick up and start working. I don't know if we want to or are easily able to integrate that into boilerplate.

Additionally, there is one other Dockerfile that is dynamically created in the catalog build: https://github.com/openshift/boilerplate/blob/994f10d6722e9035ce76ed2230a530c33e7020b6/boilerplate/openshift/golang-osd-operator/csv-generate/catalog-build.sh#L64-L86

That Dockerfile can't be updated by dependabot, and it kinda sucks we need to update boilerplate each time to pick up those image updates. I'm wondering if we can create that file as a standalone Dockerfile in the build/ directory too (like `Dockerfile.catalog`) and then update the catalog build to use it instead.

Thoughts?

dustman9000:
Thanks! I've updated PR to detect the latest ubi minimal release versions and replace the occurence of ubi-*:latest in the Dockerfile. I think we should address the catalog registry Dockerfile updates in a follow on PR.
The script snippets have been tested on an existing OSD Operator repo.
/unhold

boilerplate/openshift/golang-osd-operator/update
```diff
@@ -10,14 +10,45 @@ source $CONVENTION_ROOT/_lib/common.sh
...
+    ${SED?} -i "s,${UBI_STANDARD_IMAGE_LATEST},${UBI_STANDARD_IMAGE_RELEASE}," $DOCKERFILE
+elif grep -q ${UBI_MINIMAL_IMAGE_LATEST} $DOCKERFILE; then
+    echo "Overwriting ${DOCKERFILE}'s ${UBI_MINIMAL_IMAGE_LATEST} image to ${UBI_MINIMAL_IMAGE_RELEASE}"
+    ${SED?} -i "s,${UBI_MINIMAL_IMAGE_LATEST},${UBI_MINIMAL_IMAGE_RELEASE}," $DOCKERFILE
+elif grep -q ${UBI_MICRO_IMAGE_LATEST} $DOCKERFILE; then
+    echo "Overwriting ${DOCKERFILE}'s ${UBI_MICRO_IMAGE_LATEST} image to ${UBI_MICRO_IMAGE_RELEASE}"
+    ${SED?} -i "s,${UBI_MICRO_IMAGE_LATEST},${UBI_MICRO_IMAGE_RELEASE}," $DOCKERFILE
+else
+    echo "Did not detect any ubi latest tags in ${DOCKERFILE}, skipping replacement."
+fi
```
cblecker, line 49: Suggested improvement. This dedupes some of the logic, as well as only makes the network calls if we detect a latest tag and only for the specific image in question.

```suggestion
# Update any UBI images using latest tags to use a versioned tag that is compatible with dependabot
for ubi-latest in $(grep -oE 'registry.access.redhat.com/ubi[7-9]/ubi.*?:latest' ${DOCKERFILE}); do
    replacement_image=$(skopeo inspect --override-os linux --override-arch amd64 docker://${ubi-latest} --format "{{.Name}}:{{.Labels.version}}-{{.Labels.release}}")
    echo "Overwriting ${DOCKERFILE}'s ${ubi-latest} image to ${replacement_image}"
    ${SED?} -i "s,${ubi-latest},${replacement_image}," ${DOCKERFILE}
done
```

dustman9000: Much better! There were some issues with your suggestion I fixed after local testing.

boilerplate/openshift/golang-osd-operator/update
```diff
@@ -10,14 +10,45 @@ source $CONVENTION_ROOT/_lib/common.sh
...
 
+# Add codecov configuration
 echo "Copying .codecov.yml to your repository root."
 cp ${HERE}/.codecov.yml $REPO_ROOT
 
-# TODO: boilerplate more of Dockerfile
+# Add dependabot configuration
+mkdir -p $REPO_ROOT/.github
+if [ ! -f ${REPO_ROOT}/.github/dependabot.yml ]; then
+    # Only copy dependabot config if it doesn't exist (to allow for custom configurations)
```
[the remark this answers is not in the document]
dustman9000: Good point, updated.

PR #256, opened 2022-12-08
Add OWNER_ALIASES To golang-osd-operator Convention

dustman9000:
This adds OWNER_ALIASES to make osd operator repo assignment easier to manage by assigning SREP functional team aliases to the OWNERS file instead of individual users. This will also ease the joiner/leaver process.


2uasimojo:
This is cool... but can we do it so it's not tied to a specific convention?

dustman9000:
I only want this in the OSD Operator convention for now. No need to share this across other conventions.

mjlshen (approved):
/lgtm

This allows leaver process can be centralized to updating this repo + updating boilerplate for consumers

PR #260, opened 2023-01-11
SDCICD-854  exclude osde2e folder from unit tests 

ritmun:
[SDCICD-854](https://issues.redhat.com//browse/SDCICD-854) adding /osde2e framework to operators. This creates a /osde2e folder with e2e tests within it. 

Need to exclude this folder from unit tests.


mjlshen (approved):
/lgtm

PR #264, opened 2023-02-03
golang-osde2e convention renamed to focus on operators

ritmun:
golang-osde2e convention depends on operator config file. Renamed it to `osd-golang-operator-osde2e`  to indicate dependency and added a var check to inform consumer.


boilerplate/openshift/osd-golang-operator-osde2e/standard.mk
```diff
@@ -1,4 +1,7 @@
 # Validate variables in project.mk exist
+ifndef OPERATOR_NAME
+$(error OPERATOR_NAME is not set; only operators should consume this convention; check project.mk file)
```
2uasimojo, line 3: but but but... osde2e isn't an operator?

2uasimojo (commented):
Couple things here:
- Renaming a convention is not a trivial thing. Consumers would have to do special things to get across this commit. And I'm not sure we ever figured out what those things would be. First unsubscribe from the old name, commit, then subscribe to the new one? I'm not even sure we implemented unsubscribe. Investigation would be needed.
- The fact that the convention uses an `OPERATOR_NAME` variable doesn't mean the consumer is an operator. That's just an artifact of how this convention was born out of osd-golang-operator. If that's bugging you, I would suggest simply renaming that variable in this convention.

boilerplate/openshift/osd-golang-operator-osde2e/standard.mk
ritmun: > Couple things here:
> 
> * Renaming a convention is not a trivial thing. Consumers would have to do special things to get across this commit. And I'm not sure we ever figured out what those things would be. First unsubscribe from the old name, commit, then subscribe to the new one? I'm not even sure we implemented unsubscribe. Investigation would be needed.
> * The fact that the convention uses an `OPERATOR_NAME` variable doesn't mean the consumer is an operator. That's just an artifact of how this convention was born out of golang-osd-operator. If that's bugging you, I would suggest simply renaming that variable in this convention.

I think this will answer both comments: the osde2e convention is brand new and no consumer is using it yet. So it's trivial right now to rename it, but might not be in future.

About OPERATOR_NAME, as description states, this convention is right now designed only to work with operators. So if a non operator tries to use it, they will get appropriate error message when running related targets. The readme file is updated with that note.

Further explanation of why this renaming is done is in this thread https://redhat-internal.slack.com/archives/CMK13BP4J/p1675442696878219?thread_ts=1675291480.396139&cid=CMK13BP4J

2uasimojo: I'm sorry, I misunderstood how and by whom this convention was going to be used. I'll step out of the way :)

PR #265, opened 2023-02-06
Fixes

ritmun:
Variable named CONVENTION_DIR in osde2e convention to OSDE2E_CONVENTION_DIR. It was conflicting with CONVENTION_DIR of golang-osd-operator convention which defines lint, converage and images targets used by prow jobs. This makes prow jobs fail.

### DIFF
--- a/boilerplate/openshift/osd-golang-operator-osde2e/e2e-harness-generate.sh
+++ b/boilerplate/openshift/osd-golang-operator-osde2e/e2e-harness-generate.sh
@@ -6,14 +6,14 @@ cmd=${0##*/}
 
 usage() {
     cat <<EOF
-Usage: $0 OPERATOR_NAME CONVENTION_DIR
+Usage: $0 OPERATOR_NAME OSDE2E_CONVENTION_DIR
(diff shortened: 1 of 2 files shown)


boilerplate/openshift/osd-golang-operator-osde2e/e2e-harness-generate.sh
```diff
@@ -6,14 +6,14 @@ cmd=${0##*/}
...
-Usage: $0 OPERATOR_NAME CONVENTION_DIR
+Usage: $0 OPERATOR_NAME OSDE2E_CONVENTION_DIR
  
 EOF
     exit -1
 }
 
 OPERATOR_NAME=$1
-CONVENTION_DIR=$2
+OSDE2E_CONVENTION_DIR=$2
```
2uasimojo, line 16: nit: since this is a shell script, and it's taking this as an argument rather than from the env, it wasn't necessary to rename this instance of the variable.

2uasimojo (commented):
/lgtm

PR #267, opened 2023-02-06
removing default target from osde2e convention

ritmun:
Trying to fix fips related build error coming from `go build` from `images` presubmit prow job from operator consumer


2uasimojo:
makes sense

/lgtm

boilerplate/openshift/golang-osd-operator-osde2e/e2e-harness-generate.sh
```diff
@@ -56,7 +56,7 @@ echo "// THIS FILE IS GENERATED BY BOILERPLATE. DO NOT EDIT.
 //go:build integration
 // +build integration
 
-package osde2e_tests
+package osde2etests
```
dustman9000, line 59: Shouldn't these be separate packages?

PR #273, opened 2023-02-16
osde2e convention fix: Adding necessary "Test" prefix to e2e test runner function

ritmun:
Main test function is missing "Test" prefix. This makes `go test` not recognize the osde2e folder at all.


mjlshen (approved):
/lgtm

PR #275, opened 2023-03-02
Fixes to allow olm-registry dockerfiles to work

cblecker:
Two changes:
- First, install skopeo into the build image
- Second, don't fail validations if there is a difference in the Dockerfiles. We expect some customization here with using dependabot to update images

### DIFF
--- a/config/tag-check.sh
+++ b/config/tag-check.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -xe
+set -e
 
 # If you mess with the build image, you must publish a new
 # image-v{X}.{Y}.{Z} tag. We can't automatically generate that tag,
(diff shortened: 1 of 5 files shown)


config/tag-check.sh
```diff
@@ -23,14 +23,14 @@ fi
...
 # Since we're in a PR, and there may be multiple commits, we want to
-# check all of them; so compare against the last merge commit before
-# this one. This should work both locally (that should correspond to
-# upstream/master) and in CI (which creates a merge commit to test
-# against).
-# NOTE: This assumes we always merge with merge commits.
-last_merge_commit=$(git log --merges -n1 HEAD^ --format=%H)
-if [[ -n "$(git diff $last_merge_commit --name-only config/)" ]]; then
+# check all of them; so compare against the fork point of this branch.
+# Don't compare
```
2uasimojo, line 27: truncated comment?

cblecker: 🤦 yes. I'll tackle in a follow up as I don't want to repush the image/tag if that's okay?

2uasimojo: absolutely

PR #276, opened 2023-03-02
Bump base image to v3.0.3

cblecker:
Follow up to #275

### DIFF
--- a/.ci-operator.yaml
+++ b/.ci-operator.yaml
@@ -1,5 +1,4 @@
-# NOTE: Keep this in sync with config/Dockerfile's FROM
 build_root_image:
   namespace: openshift
   name: boilerplate
-  tag: image-v3.0.1
+  tag: image-v3.0.3


config/tag-check.sh
```diff
@@ -24,7 +24,8 @@ fi
 
 # Since we're in a PR, and there may be multiple commits, we want to
 # check all of them; so compare against the fork point of this branch.
-# Don't compare
+# Don't compare the config/Dockerfile or config/tag-check.sh files, as
+# they're not impacted by build image changes.
```
2uasimojo, line 28: I think I can get behind ignoring tag-check.sh changes at this stage of boilerplate's maturity.

However, I would say that dockerfile changes *do* impact the build image. One of the things we were trying to accomplish by standardizing the build image was predictability/consistency: so that an image at a given tag would never change. If the dockerfile gets its base image bumped, that changes the bp backing image, which warrants a new version number.

Unfortunately, this leaves us with the choice of keeping this process manual, or a making bot that can do the whole retagging thing.

I don't have a good answer here, I'm afraid.

cblecker: Right now this exact Dockerfile is *only* a FROM statement.

We could possibly make this conditional/variable input in some other script?

PR #282, opened 2023-04-06
reverting docker registry credentials for test-harness for app-interface

ritmun:
reverting changes from this morning in order to run the image publish job in app-interface instead of prow


mjlshen (approved):
/lgtm

PR #283, opened 2023-04-11
removing gomod tidy, let go test update gomod

ritmun:
Removing `go mod tidy` statement from e2e-harness-build target. `go test` should update mod file if necessary.


boilerplate/openshift/golang-osd-operator-osde2e/standard.mk
```diff
@@ -69,7 +69,6 @@ e2e-harness-generate:
 e2e-harness-build: GOFLAGS_MOD=-mod=mod
 e2e-harness-build: GOENV=GOOS=${GOOS} GOARCH=${GOARCH} CGO_ENABLED=0 GOFLAGS="${GOFLAGS_MOD}"
 e2e-harness-build:
-	go mod tidy
```
mjlshen, line 72: I'm not sure if this will work without at least a `go mod download`, which may run into the same behavior

mjlshen (changes requested):

PR #296, opened 2023-07-25
Fix bugs preventing registry from running in fips mode

mjlshen:
registry.redhat.io/openshift4/ose-operator-registry:v4.12 is the OCP released image that has more guarantees about its release process and ubi-micro needs OpenSSL to be available in order to run in fips mode. Validated that this Dockerfile produces a registry image that can run on a fips enabled cluster.

[OSD-17631](https://issues.redhat.com//browse/OSD-17631)

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/Dockerfile.olm-registry
+++ b/boilerplate/openshift/golang-osd-operator/Dockerfile.olm-registry
@@ -1,9 +1,10 @@
-FROM quay.io/openshift/origin-operator-registry:4.12 AS builder
+FROM registry.redhat.io/openshift4/ose-operator-registry:v4.12 AS builder
 ARG SAAS_OPERATOR_DIR
 COPY ${SAAS_OPERATOR_DIR} manifests
 RUN initializer --permissive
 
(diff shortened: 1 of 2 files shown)


boilerplate/openshift/golang-osd-operator/Dockerfile.olm-registry
```diff
@@ -1,9 +1,10 @@
-FROM quay.io/openshift/origin-operator-registry:4.12 AS builder
+FROM registry.redhat.io/openshift4/ose-operator-registry:v4.12 AS builder
 ARG SAAS_OPERATOR_DIR
 COPY ${SAAS_OPERATOR_DIR} manifests
 RUN initializer --permissive
 
-FROM registry.access.redhat.com/ubi8/ubi-micro:latest
+# ubi-micro does not work for clusters with fips enabled unless we make OpenSSL available
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
```
mjlshen, line 7: If we don't want to swap to ubi-minimal we could use the approach here: https://github.com/openshift/boilerplate/pull/294/files but I think we would be kind of bypassing vulnerability scanners with that approach.

PR #297, opened 2023-07-27
CGO_ENABLED=1 for FIPS compliance

mjlshen:
https://developers.redhat.com/articles/2022/05/31/your-go-application-fips-compliant

Additionally each consuming operator must use ubi-minimal, such as https://github.com/openshift/aws-vpce-operator/blob/b765a7f103a521e32a591ff5e4e64d6cd0324ec1/build/Dockerfile#L11


mjlshen (commented):
Additionally, we should add a check in boilerplate to "ensure" consuming operators use ubi-minimal8

PR #298, opened 2023-08-01
Use GOEXPERIMENT=strictfipsruntime

mjlshen:

Set `GOEXPERIMENT=strictfipsruntime`

This ensures that the binary will fail to run if running in a non-FIPS
Compliant environment when built with `FIPS_ENABLED=true`.

`GOEXPERIMENT=strictfipsruntime` is not supported by Go generally and is
something that Red Hat is supporting in our own fork. When building 
locally, developers should set `FIPS_ENABLED=false` to get around this.

Using this requires that our boilerplate base image gets rebuilt to pull in the latest `registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.19-openshift-4.12`


mjlshen:
@dustman9000 brought up that we should test that this runs on clusters that are not fips enabled + a 4.10 cluster. I have validated that operators continue to run by testing MNMO and AVO

PR #300, opened 2023-08-02
Fix bug with printing output in Makefiles

mjlshen:
Currently, this results in:

```
boilerplate/openshift/golang-osd-operator/standard.mk:97: *** missing separator.  Stop.
```

Tested the replacement syntax and confirmed it works inside the CI container


sam-nguyen7 (commented):
/lgtm

PR #304, opened 2023-08-11
golang-osd-operator: optionally generate operator manifests using kustomize

mjlshen:
This is how RMO currently generates its manifests using kustomize: https://github.com/openshift/route-monitor-operator/blob/master/Makefile#L86-L89 which is based on how this is done upstream with kubebuilder/operator-sdk https://github.com/kubernetes-sigs/kubebuilder/blob/32720bc7b8064f415b618ba820ed8f8b80f4a667/pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go#L176 

With this change, RMO should be able to be fully boilerplated without impacting existing operators.

Tested that it both does the right thing for RMO when `config/default` exists (generates the same manifests as before with no changes) and also for AVO when `config/default` does not exist (doesn't attempt to generate any manifests). This will let operators migrate to kustomize-based manifest generation at their own pace.

[OSD-15779](https://issues.redhat.com//browse/OSD-15779)

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/standard.mk
+++ b/boilerplate/openshift/golang-osd-operator/standard.mk
@@ -190,6 +190,8 @@ endef
 
 CONTROLLER_GEN = controller-gen
 OPENAPI_GEN = openapi-gen
+KUSTOMIZE = kustomize
+YQ = yq
 
 .PHONY: op-generate
 ## CRD v1beta1 is no longer supported.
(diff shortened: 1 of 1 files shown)


boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -208,8 +210,16 @@ openapi-generate:
 			-h /dev/null \
 			-r "-"
 
+.PHONY: manifests
+manifests:
+# Only use kustomize to template out manifests if the path config/default exists
+ifneq (,$(wildcard config/default))
+	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
+	$(KUSTOMIZE) build config/default | $(YQ) -s '"deploy/" + .metadata.name + "." + .kind + ".yaml"'
```
dustman9000, line 218: Can you add an else with an `@echo bla bla`? Otherwise this will break as there will be no command with the target.

mjlshen: Oh interesting, I did try this with AVO and it worked, but I do like the idea of a log anyway to encourage people to swap to kustomize and maybe I have a newer version of make or something

PR #305, opened 2023-08-16
push latest harness tag

ritmun:
Recently the latest tag on e2e test harness was replaced with operator version used in pipeline (commit hash)

Adding it back to continue testing `latest` tags on non-ci, periodic jobs, which will not refer to specific operator versions


mjlshen (approved):
/lgtm

PR #319, opened 2023-11-17
We need to stay on RHEL 8 for compatible GCC versions

mjlshen:
Since all our operators build with `CGO_ENABLED=1` for FIPS compliance, we need to remain in lockstep between this image, which builds binaries and the runtime base image.

Currently, the runtime base image must be ubi8-based also for FIPS compliance, so upgrading boilerplate results in:

```
./aws-vpce-operator: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by ./aws-vpce-operator)
./aws-vpce-operator: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by ./aws-vpce-operator)
```

[OSD-19423](https://issues.redhat.com//browse/OSD-19423)


mjlshen (commented):
/hold

PR #321, opened 2023-11-27
feat: Support multiple deployments

supreeth7:
Currently, the `golang-osd-operator` convention assumes and only supports a single deployment per operator. But there are scenarios where an operator might have more than one non-op deployment that needs to be deployed alongside. 
This PR now adds support for more than one deployment per operator for the above convention.

### Notes for reviewer:-
- This PR enhances the support for a single deployment to now support a maximum of **two** deployments per operator.
- The additional deployment is completely optional and the convention will continue to function the same if it isn't specified.
- If an additional deployment image has to be built and appended to the CSV as part of the build-push process, then the consumer needs to:
  - Specify `SupplementaryImage` which is the deployment name in the operator repository's `config/config.json`.
  - Define the image to be built as `ADDITIONAL_IMAGE_SPECS` in the consuming repository's Makefile, Boilerplate later parses this image as part of the build process; [ref](https://github.com/openshift/boilerplate/blob/master/boilerplate/openshift/golang-osd-operator/standard.mk#L56), for example:
    ```bash
    # Additional Deployment Image
    define ADDITIONAL_IMAGE_SPECS
    build/Dockerfile.webhook $(SUPPLEMENTARY_IMAGE_URI)
    end
    ```
  -  The `SUPPLEMENTARY_IMAGE_URI` is generated via Boilerplate [here](https://github.com/openshift/boilerplate/compare/master...supreeth7:boilerplate-upstream:multi-deploy?expand=1#diff-aef0082c5bb612084a57562263d822bb2a2a089ec35cfcf3feec179673c56b11R59).
  - Modify the CSV in `config/csv-template.yaml` in the consuming repository to contain an additional deployment name.

### Testing
- I've tested the changes locally via the AppSRE build pipelines following https://github.com/openshift/boilerplate/blob/master/boilerplate/openshift/golang-osd-operator/app-sre.md.
- The OLM bundle created off these changes: https://gitlab.cee.redhat.com/sbasabat/saas-addon-operator-bundle/-/tree/staging/addon-operator/0.1.1342-3a54ff1
- Images:
  - Operator image: https://quay.io/repository/sbasabat_openshift/addon-operator
  - Additional deployment image: https://quay.io/repository/sbasabat_openshift/addon-operator-webhook
  - Catalog image: https://quay.io/repository/sbasabat_openshift/addon-operator-registry


supreeth7:
If accepted, resolves https://issues.redhat.com/browse/OSD-19741

bdematte:
Wouldn't it make more sense to support a set of deployments (through a single parameter) rather than adding a new parameter for each additional deployment ?

supreeth7:
@bdematte - That was infact the initial plan but I do not see any operator going beyond a two-deployment model. The specific use-case behind this PR is to deploy a webhook which will soon be obsolete with the introduction of CEL, OPA, etc.

Hence pretty much locks this convention to 2 deployments avoiding a slightly more complex code.

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
```diff
@@ -61,6 +64,17 @@ if [[ -z "$IMAGE_DIGEST" ]]; then
 fi
 REPO_DIGEST=${operator_image}@${IMAGE_DIGEST}
 
+# Given a supplementary image is specified,
+# generate the image digest.
+if [ ! -z supplementary_image ]; then
+    IMAGE_DIGEST=$(skopeo inspect docker://${supplementary_image}:v${operator_version} | jq -r .Digest)
```
2uasimojo, line 70: As written we're assuming both images (the primary and the supplementary) get built every time, such that they have matching image tags. I didn't go look at the mechanics of the build process, but it seems like, in almost all cases, one or the other will be identical to the previous version.

[Later] I did go and refresh my memory on the build process. I now remember writing the support for additional images. And I thought through possible alternatives and none seem worth the trouble.

So nothing to act on here, but I'll just express mild sadness at the inefficiency of producing and storing an effectively identical image with every PR.

2uasimojo (commented):
This is cool, a pretty unobtrusive change to get a significant boost in functionality. I dig it.

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
```diff
@@ -61,6 +64,17 @@ if [[ -z "$IMAGE_DIGEST" ]]; then
 fi
 REPO_DIGEST=${operator_image}@${IMAGE_DIGEST}
 
+# Given a supplementary image is specified,
+# generate the image digest.
+if [ ! -z supplementary_image ]; then
```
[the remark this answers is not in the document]
supreeth7: Ah, thanks, bash always gets me :face_exhaling:

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
supreeth7: I can resonate with your mild sadness, it would indeed need a partial overhaul of the image building pipeline which I reckon would be a reach to support a simple functionality such as this.

supreeth7:
@2uasimojo - Thank you for reviewing!
I've made the asked changes and tested things out locally and everything checks out.
Mind giving it another look :crossed_fingers:

2uasimojo:
/lgtm

Wouldn't mind if @bdematte gave the final ack here.

bdematte:
/unhold
handling of a set of image doesn't deserve to block ongoing activities

PR #322, opened 2023-11-29
Added env var to osde2e pod template

ritmun:
Follow up change to osde2e update https://github.com/openshift/osde2e/pull/2244

Provides LOG_BUCKET env var to osde2e pod template in operator tekton jobs.


mjlshen (approved):
/lgtm

PR #323, opened 2023-11-30
feat: Introduce OLM SkipRange to CSV builds

supreeth7:
### What
- This PR aims to add the OLM [SkipRange](https://v0-18-z.olm.operatorframework.io/docs/concepts/olm-architecture/operator-catalog/creating-an-update-graph/#skiprange) functionality to Boilerplate CSV builds.
- Also addresses a critical bug in the multi-deployment builds.
  - [csv-generate.mk](https://github.com/openshift/boilerplate/blob/master/boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.mk#L3) expects `SUPPLEMENTARY_IMAGE` parameter which we are initializing to `${IMAGE_ORG}/${IMAGE_REPOSITORY}/${SUPPLEMENTARY_IMAGE_NAME}` regardless if `SUPPLEMENTARY_IMAGE_NAME` is defined in the operator config or not. This PR sets `SUPPLEMENTARY_IMAGE` to an empty string if the supplementary image name isn't explicitly specified in the operator's `config/config.go`.

### Why
- OLM currently doesn't support cross-catalog upgrades.
- Boilerplate's `golang-osd-operator` convention standardizes the catalog repositories to adhere to the naming convention`${OPERATOR_NAME}-registry`.
- Once deployed, for an existing operator to upgrade to the new Boilerplate-deployed operator which refers to the new catalog registry with `staging` & `production` channels, OLM needs to support cross-catalog upgrades.
- Cross catalog upgrades are only possible via the `olm.skipRange` annotation in the CSV.
- This will allow already deployed OSD operators that subscribe to Boilerplate in the future to have an OLM upgrade path.

#### Example Scenario
- Currently, we have the  OSD Addon Operator adopting Boilerplate and it's catalog image points to `addon-operator-index` of the `stable` channel; while post Boilerplate adoption it now points to `addon-operator-registry` of `staging/production` channel.
- Once deployed, for this operator to have an upgrade path between it's old and the new catalog, we need to append the OLM SkipRange annotation.

### Testing
- [x] Tested the build process locally with skiprange enabled/disabled.


mjlshen (approved):
/lgtm

PR #325, opened 2024-01-16
add SRE group leads to owners aliases

maorfr:
follow up on https://github.com/openshift/managed-cluster-config/pull/1841

to enable group leads to become owners in a streamlined way


mjlshen (approved):
/lgtm

PR #335, opened 2024-02-13
Fix build push

typeid:
**What?**
 This PR changes the build-push script (specific to boilerplate) to create the image for the most recent tag if it doesn't exist yet. 

**Why?**
Currently, boilerplate assumes the latest commit on master is the one a tag is created for. In case a commit is merged between the creation of a new PR to create a new boilerplate version and the merge, this will break the image creation.


mjlshen (approved):
/lgtm

PR #336, opened 2024-02-13
Use existing container engine instead of podman

typeid:
Fixes the previous PR that assumed podman as existing container engine:
https://github.com/openshift/boilerplate/pull/335


mjlshen (approved):
/lgtm

PR #339, opened 2024-02-19
Update golangci lint version

ritmun:
This version is very old. Updated to match the one in [golang-lint convention](https://github.com/openshift/boilerplate/blob/master/boilerplate/openshift/golang-lint/ensure.sh#L8)


2uasimojo:
Discussed with @ritmun, who is going to run this against a consuming repo or two to make sure they don't splode and block subsequent bp updates.

PR #342, opened 2024-03-19
golang-osd-operator: use new path to TEMPLATE_FILE

jbpratt:
this file will not exist in most repositories yet, it will need to be added during the migration process


boilerplate/openshift/golang-osd-operator/standard.mk
mjlshen, line 1: @jbpratt this seems like a benign change, but do you have a card describing the larger effort around this? Thanks!

jbpratt: hey @mjlshen , I've had to rework some stuff to work towards running this in a pipeline and walking through the initial migration from SSS to PKO. I just opened SDCICD-1243 to track this effort. More details can be found in SD-DDR-0022!

mjlshen: Yup, thanks, just leaving breadcrumbs for future visitors :D

mjlshen (approved):
/lgtm

PR #343, opened 2024-03-29
Update push steps to use containerized skopeo

mjlshen:
Currently, we're depending on the version of skopeo installed on a Jenkins VM, which has gone out of date. Yes, the VM should be updated, but this gives us more control over the version of skopeo being used in CI.

[OSD-21843](https://issues.redhat.com//browse/OSD-21843)


2uasimojo:
Unrelated to this PR, [this](https://github.com/openshift/boilerplate/blob/d11572e7fe85fd5c236d430c2d047ddf5b3869c3/config/build_image-v5.0.0.sh#L42-L43) is biting ya :(

Suggest nailing that down to a specific version. Not sure why we decided not to do that originally.

mjlshen:
Caused by https://github.com/kubernetes-sigs/controller-runtime/issues/2744

PR #345, opened 2024-04-02
[OSD-21903] Do not create a bundle that replaces itself.

bergmannf:
In certain cases it can happen that this tries to create a bundle that replaces itself, which will lead to installation failures.

As generally commits that create the same version should not be allowed csv-generate.sh checking for reverts, it should be safe to simply error out, as this seems to mostly happen, when an existing bundle gets updated with the incorrect replace, after it initially had the correct 'replaces' statement.

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
+++ b/boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
@@ -165,6 +165,15 @@ OPERATOR_PREV_VERSION=$(ls "$BUNDLE_DIR" | sort -t . -k 3 -g | tail -n 1)
 OPERATOR_NEW_VERSION="${operator_version}"
 OUTPUT_DIR=${BUNDLE_DIR}
 
+VERSION_DIR="${OUTPUT_DIR}/${OPERATOR_NEW_VERSION}"
+
+# Check if the VERSION_DIR already exists and is not empty - if so skip building
+# anything, as only timestamps would be changed.
+if [[ -d "${VERSION_DIR}" && -n $(ls -A "${VERSION_DIR}") ]]; then
+    echo "Output directory for bundle already exists and is not empty: ${VERSION_DIR}. Skipping bundle creation."
+    exit 1
+fi
+
 # If setting up a new SaaS repo, there is no previous version when building a bundle
 # Optionally pass it to the bundle generator in that case.
 if [[ -z "${OPERATOR_PREV_VERSION}" ]]; then
--- a/boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
+++ b/boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py
@@ -58,6 +58,15 @@ skip_range_enabled=args.skip_range_enabled
 
 hasMultipleDeployments = False
 
+# Verify that this is not creating a cycle.
+if prev_version == full_version:
+    print(
+        "Previous version equals current version. "
+        "This would result in a cycle that can not be installed - aborting."
+    )
+    sys.exit(1)
+
+
 class UnsupportedRegistryResourceKind(Exception):
     def __init__(self, kind, path):
         super().__init__(


2uasimojo:
Drive-by review: Can we instead/also identify the circumstances that lead to this happening? Doesn't seem like it should be possible.

bergmannf:
The root-cause seems to be failing CI builds of the bundle files that already manage to create the bundle file, but on consecutive runs replace the replaces with itself, because the last commit is the commit that created the bundle, not the previous one.

So I think either the 'get the previous commit' logic has to change, or the 'assume this did work previously and just do nothing then'.

This PR is the 'it worked before, do nothing' way, but the other way of making sure that the `previous_version` is really the previous version should fix it as well - IMO the code can not just assume the all CI runs will work, which kind of is what is happening right now AFAIU.

2uasimojo:
What I don't understand is: pushing the bundle image seems to be the last thing that happens in the pipeline. If that fails, the bundle doesn't exist, and the subsequent attempt should be okay. I guess my point is that I'm still not seeing the path whereby we can end up in this state. Do you have an example of a build of a consuming repo where this happened?

BTW, please don't let my review hold up this PR. I would rather not be the one to give it final approval in any case, since I'm no longer an active user of boilerplate.

2uasimojo:
Had an offline debug session with @bergmannf. We believe we've tracked down how this situation can arise:
- app-sre-build-deploy.sh runs csv-build, catalog-build, catalog-publish IFF the bundle image doesn't exist already
- csv-build is what puts a new commit into the saas bundle repo
- catalog-publish is what pushes a new quay image for that bundle

So if the former succeeds but the latter fails, that is how we can get into this situation:
- the quay image doesn't exist so the subsequent build thinks it has to start over, but
- the saas bundle repo is populated so the `replaces` version discoverer thingy sees the existing bundle as the latest
- ...and then I *think* we're regenerating bundle files that already exist! so when it comes time to create the commit, instead of it being a dir full of new files, it's just a delta with the new (bogus) `replaces` and a new timestamp.

So I think what we may actually want to do in this PR is have csv-generate check for the existence of the bundle dir and completely no-op. That would make that step idempotent and it would fit nicely into the workflow.

Because as written what I think will happen is we'll end up rewriting all the files, but the only delta would end up being that one timestamp. Which would be okay, but unnecessary. (Unless we think it is desirable to have that timestamp gel with the rest of the build -- but actually I think it would be better to leave it alone so we can tell when the bundle was _originally_ created.)

bergmannf:
@2uasimojo thanks for the investigation of the cause - agreed that it's the Make targets just all being run in succession with no logic depending on the previous step being successful or not.

I've updated the csv generation to exit out if a bundle directory that is not empty is encountered.

I ran this for a local operator build to verify it does what is expected - so initially it will output:

```
Bundling saas-aws-account-operator-bundle/aws-account-operator/0.1.1675-5e01ade/CustomResourceDefinition_accountclaims.aws.managed.openshift.io.yaml
Bundling saas-aws-account-operator-bundle/aws-account-operator/0.1.1675-5e01ade/CustomResourceDefinition_accountpools.aws.managed.openshift.io.yaml
Bundling saas-aws-account-operator-bundle/aws-account-operator/0.1.1675-5e01ade/CustomResourceDefinition_awsfederatedaccountaccesses.aws.managed.openshift.io.yaml
Bundling saas-aws-account-operator-bundle/aws-account-operator/0.1.1675-5e01ade/CustomResourceDefinition_awsfederatedroles.aws.managed.openshift.io.yaml
Bundling saas-aws-account-operator-bundle/aws-account-operator/0.1.1675-5e01ade/CustomResourceDefinition_accounts.aws.managed.openshift.io.yaml
Bundling saas-aws-account-operator-bundle/aws-account-operator/0.1.1675-5e01ade/Role_aws-account-operator_prometheus-k8s.yaml
Bundling saas-aws-account-operator-bundle/aws-account-operator/0.1.1675-5e01ade/RoleBinding_aws-account-operator_prometheus-k8s.yaml
Wrote ClusterServiceVersion: saas-aws-account-operator-bundle/aws-account-operator/0.1.1675-5e01ade/aws-account-operator.v0.1.1675-5e01ade.clusterserviceversion.yaml
```

When rerunning it (without cleaning up the bundle directory) it will then exit out with the following message:

```
Output directory for bundle already exists and is not empty: saas-aws-account-operator-bundle/aws-account-operator//0.1.1675-5e01ade. Skipping bundle creation.
```

mjlshen (approved):
/lgtm

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
```diff
@@ -165,6 +165,15 @@ OPERATOR_PREV_VERSION=$(ls "$BUNDLE_DIR" | sort -t . -k 3 -g | tail -n 1)
 OPERATOR_NEW_VERSION="${operator_version}"
 OUTPUT_DIR=${BUNDLE_DIR}
 
+VERSION_DIR="${OUTPUT_DIR}/${OPERATOR_NEW_VERSION}"
+
+# Check if the VERSION_DIR already exists and is not empty - if so skip building
+# anything, as only timestamps would be changed.
+if [[ -d "${VERSION_DIR}" && -n $(ls -A "${VERSION_DIR}") ]]; then
+    echo "Output directory for bundle already exists and is not empty: ${VERSION_DIR}. Skipping bundle creation."
+    exit 1
```
2uasimojo, line 174: Should this actually exit 0? As in "nothing to do" rather than "error"? IIUC exiting nonzero will fail the pipeline completely, rather than letting it fall through to the next step, which is what we want.

2uasimojo (commented):
Sorry I lost track of this one! Possible correction inline...

PR #349, opened 2024-05-13
Update ose-operator-registry

feichashao:
Update the ose-operator-registry version so that it includes the vulnerabilities fixes.

Tested with `podman build -f Dockerfile.olm-registry` in `boilerplate/openshift/golang-osd-operator`.

Not sure how should I test further.


2uasimojo:
/approve

This seems sane to me. But it would be good to have @cblecker look, as he was the origin of [this comment](https://github.com/openshift/boilerplate/blob/9abcd43fa6cc07b804084ef61b5e225c3c27f105/boilerplate/openshift/golang-osd-operator/dependabot.yml#L14) (via #274), which means he must have had some idea about how we should upgrade this registry.

/assign @cblecker

cblecker:
/lgtm

Yes, this is meant to be bumped in boilerplate -- the UBI image used in the final image is bumped locally in each repo, but the build container should be bumped here!

PR #351, opened 2024-05-16
golang-osd-operator: additional updates

jbpratt:
feedback from #350, make it obvious the these exits are a success case
and add a comment explaining the string manipulation

also including a fix to create and push the branch if it doesn't already exist

Signed-off-by: Brady Pratt <bpratt@redhat.com>

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
+++ b/boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
@@ -82,6 +82,8 @@ BUNDLE_DIR="$SAAS_OPERATOR_DIR/${operator_name}/"
 rm -rf "$SAAS_OPERATOR_DIR"
 BRANCH="$operator_channel"
 if [[ "${RELEASE_BRANCHED_BUILDS}" ]]; then
+    # operator version will be set to `X.Y.BUILD_NUMBER-commit sha`, this will
+    # be `release-X.Y`
     BRANCH="release-${operator_version%.*}"
 fi
 git clone --branch "${BRANCH}" ${GIT_PATH} "$SAAS_OPERATOR_DIR"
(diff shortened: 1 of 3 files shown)


jbpratt:
related to SDCICD-1290

jbpratt:
@2uasimojo I looked at reducing the duplication suggested https://github.com/openshift/boilerplate/pull/350#discussion_r1601860059 and https://github.com/openshift/boilerplate/pull/350#discussion_r1601878580 but I think both cases end up more complex. The names of channels and including commit-shas make it difficult since we can't just loop over a flat array.

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.sh
```diff
@@ -82,9 +82,18 @@ BUNDLE_DIR="$SAAS_OPERATOR_DIR/${operator_name}/"
...
 if [[ "${RELEASE_BRANCHED_BUILDS}" ]]; then
+    # operator version will be set to `X.Y.BUILD_NUMBER-commit sha`, this will
+    # be `release-X.Y`
     BRANCH="release-${operator_version%.*}"
 fi
-git clone --branch "${BRANCH}" ${GIT_PATH} "$SAAS_OPERATOR_DIR"
+
+git clone ${GIT_PATH} "$SAAS_OPERATOR_DIR"
+pushd "${SAAS_OPERATOR_DIR}"
+# if branch doesn't exist, checkout a new branch based on staging
```
jbpratt, line 92: The versioned branch in the saas repos needs to be created initially

2uasimojo:
> @2uasimojo I looked at reducing the duplication suggested [#350 (comment)](https://github.com/openshift/boilerplate/pull/350#discussion_r1601860059) and [#350 (comment)](https://github.com/openshift/boilerplate/pull/350#discussion_r1601878580) but I think both cases end up more complex. The names of channels and including commit-shas make it difficult since we can't just loop over a flat array. Let me know if you have any suggestions!

It might be something like #352? Feel free to use it if you can.

2uasimojo:
/lgtm

We can fup with the DRYing if desired.

PR #360, opened 2024-06-11
golang-osd-operator: checkout branchs and check remote

jbpratt:
We weren't doing a `git pull --all` so the check on L94 was not working correctly. Use `ls-remote` to check instead and create


mjlshen (approved):
/lgtm

PR #363, opened 2024-06-22
Create boilerplate v5.0.1 to support Go 1.22

mjlshen:
Breaking change: In the past we would have done a major version (e.g. v6.0.0), but the past few bumps have been painless and the Go promise of backwards support has been paying dividends. Has the side benefit of allowing the changes in `config/build_image-v5.0.0.sh` to be sanely reviewed instead of copy+pasting it and modifying versions at the same time.

Otherwise, updates the base image to support Go 1.22 for when people need it.

[OSD-24219](https://issues.redhat.com//browse/OSD-24219)


mjlshen:
/hold

For this commit to be tagged with v5.0.1 if it looks ok before merging

typeid (approved):
/lgtm

We might want to rename `config/build_image-v5.0.0.sh` in the future to e.g. `config/build_image-v5.x.x.sh`

typeid:
@mjlshen PR title should be `to support Go1.22`

mjlshen:
> We might want to rename config/build_image-v5.0.0.sh in the future to e.g. config/build_image-v5.x.x.sh

Makes sense, and we can delete the old ones too and keep renaming the current one? Or should we just add a comment in the shell script that holds the version instead? The latter seems easier to me, but I think both are better than current state

typeid:
@mjlshen I think a comment makes most sense.

mjlshen:
/assign @dustman9000 

When you have a chance could you tag the HEAD of this PR (`b4498b34f115e85864e0642456cbf65fcbd0d3e7`) with `image-v5.0.1` please?

PR #365, opened 2024-06-25
Give RVMO code path proper branch/version detection

mmazur:
Release branched builds can only work correctly if they're being run from a branch named `release-4.Y`.

This gets complicated in two places:
* If a developer wants to locally build from their feature branch, which definitively isn't named `release-4.Y`. This is taken care of by being able to run `RELEASE_BRANCH=release-4.16 make …` to override current branch detection.
* Appiface Jenkins just doesn't provide the build branch to the environment ([see](https://ci.int.devshift.net/view/osde2e/job/openshift-osd-example-operator-gh-build-release-4.16/74/injectedEnvVars/)) and the project's git repo is checked out as a detached HEAD commit, so there's no "what's my branch" information in there either. Best I can do right now is to try to extract the branch from JOB_NAME, which is what this code does… for now.

On the last point, there should be `GIT_LOCAL_BRANCH` available in the env, but for some reason it's not. I'm talking to appsre about this and if we ever get that working, I'll remove the current JOB_NAME hack and just look at `GIT_LOCAL_BRANCH`. But for now, this is the only way I can get this to work in prod.

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/project.mk
+++ b/boilerplate/openshift/golang-osd-operator/project.mk
@@ -17,9 +17,30 @@ VERSION_MAJOR?=0
 VERSION_MINOR?=1
 
 ifdef RELEASE_BRANCHED_BUILDS
-    BRANCH_NAME := $(shell git rev-parse --abbrev-ref HEAD | grep -E '^release-[0-9]+\.[0-9]+$$')
-    ifneq ($(BRANCH_NAME),)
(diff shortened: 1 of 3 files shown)


jbpratt:
/lgtm

needs approval from boilerplate approvers

mmazur:
@2uasimojo @mjlshen hi, can I get a review?

mjlshen (approved):
/lgtm

PR #366, opened 2024-06-27
Take container auth file from jenkins

mmazur:
In July all osd operators will be built in a rhel8 jenkins node. And that was causing a container registry access error, at least for osd-example-operator.

This change fixes the issue (at least for osd-example-operator). It's based on [this change to Hive](https://github.com/openshift/hive/pull/2321/files).


mjlshen (approved):
/lgtm

PR #369, opened 2024-07-05
Final final rvmo changes (v2)

mmazur:
- Switch to rvmo template file dir, not a single file
- Extract actual REPO_NAME and pass to rvmo script


mjlshen (approved):
/lgtm

PR #373, opened 2024-07-09
[OSD-13681] Added a prefix 'g' to handle leading zeroes in CSV version

aliceh:
In this commit, I added a prefix of 'g' to the hash to handle leading zeroes in CSV version, which was breaking OLM. 
This is a convention used in other places. For example:
machine-api-operator ran into this https://bugzilla.redhat.com/show_bug.cgi?id=2015188 and fixed this by prepending a "g" for git I guess https://github.com/openshift/doozer/pull/527. 

https://issues.redhat.com/browse/OSD-13681

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/standard.mk
+++ b/boilerplate/openshift/golang-osd-operator/standard.mk
@@ -42,7 +42,7 @@ endif
 
 # Generate version and tag information from inputs
 COMMIT_NUMBER=$(shell git rev-list `git rev-list --parents HEAD | grep -E "^[a-f0-9]{40}$$"`..HEAD --count)
-CURRENT_COMMIT=$(shell git rev-parse --short=7 HEAD)
+CURRENT_COMMIT=g$(shell git rev-parse --short=7 HEAD)
 OPERATOR_VERSION=$(VERSION_MAJOR).$(VERSION_MINOR).$(COMMIT_NUMBER)-$(CURRENT_COMMIT)
 
 OPERATOR_IMAGE=$(IMAGE_REGISTRY)/$(IMAGE_REPOSITORY)/$(IMAGE_NAME)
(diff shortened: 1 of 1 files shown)


2uasimojo:
Nice plan! This strategy may also fix the situation where a short hash happens to be all numeric and is thus interpreted as a number rather than a string. Not sure if this delta gets all the way down there.

boilerplate/openshift/golang-osd-operator/standard.mk
```diff
@@ -42,7 +42,7 @@ endif
 
 # Generate version and tag information from inputs
 COMMIT_NUMBER=$(shell git rev-list `git rev-list --parents HEAD | grep -E "^[a-f0-9]{40}$$"`..HEAD --count)
-CURRENT_COMMIT=$(shell git rev-parse --short=7 HEAD)
+CURRENT_COMMIT=g$(shell git rev-parse --short=7 HEAD)
 OPERATOR_VERSION=$(VERSION_MAJOR).$(VERSION_MINOR).$(COMMIT_NUMBER)-$(CURRENT_COMMIT)
```
mjlshen, line 46:
```suggestion
CURRENT_COMMIT=$(shell git rev-parse --short=7 HEAD)
OPERATOR_VERSION="$(VERSION_MAJOR).$(VERSION_MINOR).$(COMMIT_NUMBER)-g$(CURRENT_COMMIT)"
```
I think we should move the `g` down to `OPERATOR_VERSION` just because the `CURRENT_COMMIT` variable is used elsewhere in this file

mjlshen (changes requested):

boilerplate/openshift/golang-osd-operator/standard.mk
aliceh: Makes sense, thanks!

mjlshen (approved):
/lgtm

mjlshen (commented):

PR #378, opened 2024-07-24
KONFLUX-3729: add containerfile for konflux

jbpratt:
the base image needs to be one built in Konflux, the brew golang builder works for now.

https://github.com/openshift/osd-example-operator/pull/290/checks?check_run_id=27879765416

Above is a test build of the image to show it builds correctly


mjlshen (approved):
/lgtm

PR #380, opened 2024-07-25
fix the python usage in build-opm-catalog.sh too

tremes:


2uasimojo (commented):
/lgtm

three step iterations unfortunately

PR #381, opened 2024-07-29
add check that new version and previous version are not the same.

Mhodesty:
https://issues.redhat.com/browse/OSD-14964


boilerplate/openshift/golang-osd-operator/csv-generate/catalog-publish.sh
```diff
@@ -41,6 +41,11 @@ BUNDLE_DIR="${SAAS_OPERATOR_DIR}/${operator_name}"
 OPERATOR_NEW_VERSION=$(ls "${BUNDLE_DIR}" | sort -t . -k 3 -g | tail -n 1)
 OPERATOR_PREV_VERSION=$(ls "${BUNDLE_DIR}" | sort -t . -k 3 -g | tail -n 2 | head -n 1)
 
+if [[ "$OPERATOR_NEW_VERSION" == "$OPERATOR_PREV_VERSION" ]]; then
+    echo "New version and previous version are identical. Exiting."
+    exit 1
```
mjlshen, line 46:
```suggestion
    exit 0
```
Depending on the use-case, do we want to `exit 0` here as a success? Or should this be considered a CI failure?

Mhodesty: This should be considered a failure. This will be triggered if the first part of the pipeline doesn't run correctly.

mjlshen (approved):
/lgtm

PR #386, opened 2024-07-31
fix the opm-build-push target after d4f5e8add3d7afdd361c069d18ae758294…

tremes:
…3273bd commit

The https://github.com/openshift/boilerplate/commit/d4f5e8add3d7afdd361c069d18ae7582943273bd broke the `opm-build-push`, because we don't have any catalog image with this `g` prefix so this is basically temporary workaround to fix that.


mjlshen (approved):
/lgtm

PR #389, opened 2024-08-01
BUGFIX: Typo while accessing OLM_CATALOG_IMAGE env var

ncaak:
### summary

* Fixed a typo when accessing an env var that caused the image to contain curly braces and thus fail to build.


2uasimojo (commented):
Oops 😁

/lgtm

PR #390, opened 2024-08-07
Fix repo image tag prefix

ncaak:
### summary

* Fixed introduced change on repo images' tag prefix, as it breaks upstream builds


ncaak:
Sorry, I'll need a bit of time to clarify one aspect of this change.

fahlmant (commented):
/hold
The addition of the `g` is purposeful. Can you give some more detail on why this is being removed?

PR #397, opened 2024-08-19
OSD-25018 - Address opm bug

theautoroboto:
This PR will address a documented bug with opm.

- Addressing an opm bug where the error "Error: Invalid bundle xxxx.vx.x.x, replaces nonexistent bundle xxxx.vx.x.x"
- Updating opm version


mjlshen (approved):
/lgtm

PR #399, opened 2024-08-26
Add Cicada team to OWNER_ALIASES

mmazur:
Cicada was the one team missing from the owners file despite being co-responsible for parts of and regularly contributing to a variety of srep-owned repos. And this commingling is going to increase in the future with rvmo, progressive delivery, konflux and the like.

Was conflicted on how to name it (Cicada is two reorgs away from being an actual functional team in srep, so I couldn't name it srep-functional-team-X), so I went with a "virtual" name that describes what the people are supposed to do, to make sure it's future-proof. If anybody got a better name, please suggest it.


mjlshen (approved):
/lgtm

PR #401, opened 2024-09-03
Update Team Aurora roster

abyrne55:
Alex and Raf have moved on to other roles within Red Hat, so this PR removes them from the Team Aurora group


mjlshen (approved):
/lgtm

PR #415, opened 2024-09-30
Updated github cli version to fix vulnerability.

Mhodesty:


mjlshen (commented):
@Mhodesty The good news is that CI is fixed. The bad news is that you'll need to rebase your PR and make the associated changes to `config/Dockerfile` and `config/Dockerfile.konflux` now...and we may as well go to gh CLI version `2.58.0` as it's been released in the meantime :D

PR #417, opened 2024-10-03
Build boilerplate images with Dockerfile primitives

mjlshen:
We've been maintaining `build_image.sh` as a TODO for a while and it prevents container builds from having their build layers cached properly. In addition, this allows us to standardize the Dockerfile between OCP CI and AppSRE. Konflux is still slightly different for now, but we're anticipating that the OCP CI and AppSRE images will no longer be needed in the near future.

[OSD-25819](https://issues.redhat.com//browse/OSD-25819)


.tekton/image-pull-request.yaml
mjlshen, line 1: This change is to mirror the higher memory limits we have set in Prow https://github.com/openshift/release/blob/00ecd90667e62ef3b718ddaa60727db92fcfbf75/ci-operator/config/openshift/boilerplate/openshift-boilerplate-master.yaml#L11

PR #420, opened 2024-10-09
Reconcile RHEL versions and update build paths.

Mhodesty:
The image path at registry.ci.openshift.org/ocp/builder is causing issues. Updating this to registry.ci.openshift.org/openshift/release to try to remedy this.

Initially the build was also updated to change images from rhel-8 to rhel-9, but this introduced more issues in testing. I've removed that part of this PR and will continue troubleshooting. The rhel-8 to rhel-9 upgrade will come in a later PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -405,7 +405,7 @@ it's not available in boilerplate:image-v2.Y.Z
 
 1. Update config/Dockerfile
     ```
-    FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.19-openshift-4.12
+    FROM registry.ci.openshift.org/openshift/release:rhel-9-release-golang-1.22-openshift-4.17
     ```
 2. Then, update the rest of boilerplate accordingly, push a new tag, and mirror the image into openshift/release
 to create boilerplate:image-v3.0.0
--- a/README.md
+++ b/README.md
@@ -421,7 +421,7 @@ it's not available in boilerplate:image-v2.Y.Z
 
 1. Update config/Dockerfile
     ```
-    FROM registry.ci.openshift.org/openshift/release:rhel-9-release-golang-1.22-openshift-4.17
+    FROM registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.22-openshift-4.17
     ```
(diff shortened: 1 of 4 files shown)


config/Dockerfile
mjlshen, line 1: `ARG BASE_IMAGE` exists here because the base images used to be different between when this Dockerfile was used in Prow and when it was used in app-interface. Now that they are the same, we should remove this ARG + comment + corresponding pieces of `config/app-sre-build-push.sh` imo.

```
FROM registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.22-openshift-4.17 AS builder
...
FROM registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.22-openshift-4.17
```

README.md
```diff
@@ -421,7 +421,7 @@ it's not available in boilerplate:image-v2.Y.Z
 
 1. Update config/Dockerfile
     ```
-    FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.19-openshift-4.12
+    FROM registry.ci.openshift.org/openshift/release:rhel-9-release-golang-1.22-openshift-4.17
```
mjlshen, line 424:
```suggestion
    FROM registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.22-openshift-4.17
```

config/Dockerfile.konflux
mjlshen, line 1: Similar to above, I would revert these changes as we don't want to change the base image at runtime

mjlshen (changes requested):

config/Dockerfile
Mhodesty: Could you explain this a little bit further? I don't understand the application here in the sense that we're just updating the versioning.

mjlshen: Oh I mean that we are no longer overriding the base image in `config/app-sre-build-push.sh`

So we can just explicitly set the base image (without `ARG`), remove this comment
```
# When updating this default BASE_IMAGE, remember to also update it in app-sre-build-push.sh
```
and remove the override logic in `config/app-sre-build-push.sh`

mjlshen (approved):
/lgtm

PR #488, opened 2025-02-04
SDE-4270 Switch OSDe2e to use OCM Client Credentials

MrSantamaria:
https://issues.redhat.com/browse/SDE-4270

Follow up to: 
https://github.com/openshift/osde2e/pull/2822/files

This change adds the OCM Client Credentials that OSDe2e will start using once offline tokens are deprecated.


joshbranham:
hey @MrSantamaria just confirming, is this PR good to go?

PR #492, opened 2025-02-07
OSD-28159 - Enable OLM Skip Range

anispate:
Configure boilerplate to generate an OLM bundle with a skip range pointing to the latest version. This will allow OLM to upgrade past broken replaces chains in which versions get abadoned without a path forward. The new OLM bundle will contain a single version, referencing the latest build with a skipRange to the version. Part of [OSD-28159](https://issues.redhat.com//browse/OSD-28159) ticket.

### DIFF
--- a/test/projects/file-generate/config/config.go
+++ b/test/projects/file-generate/config/config.go
@@ -15,6 +15,7 @@
 package config
 
 const (
-	OperatorName      string = "file-generate"
-	OperatorNamespace string = "file-generate"
+	OperatorName       string = "file-generate"
+	OperatorNamespace  string = "file-generate"
+	EnableOLMSkipRange        = "true"
 )


test/projects/file-generate/config/config.go
```diff
@@ -15,6 +15,7 @@
 package config
 
 const (
-	OperatorName      string = "file-generate"
-	OperatorNamespace string = "file-generate"
+	OperatorName       string = "file-generate"
+	OperatorNamespace  string = "file-generate"
+	EnableOLMSkipRange        = "true"
```
joshbranham, line 20: `operator-sdk` doesn't generate this as an option, rather we embed it after the fact on a per-project basis and then read this file to generate the CSV.

/close

PR #510, opened 2025-03-24
[OSD-28770] and [OSD-28769] We are cleaning up some aliases for our t…

pheckenlWork:
We are just removing some aliases from our team for users who have moved to other teams.


joshbranham (commented):
/lgtm

PR #515, opened 2025-04-04
fix: match operator name var more strictly

jbpratt:
current expression matches a variable in MUO causing the files to be generated incorrectly. This matches more strictly for the OperatorName variable


dustman9000:
Can you please add a PR description?

jbpratt:
> Can you please add a PR description?

:) update PTAL

PR #517, opened 2025-04-08
OSD-29370: Configure `HOME` for golang-lint and golang-codecov boilerplate conventions

joshbranham:
In https://github.com/openshift/boilerplate/commit/1e947d2b7daee575dbc0283c647b9945a8081c8f we added the below `HOME` override to ensure there is a place for `.cache` to be created with permissions in prow.

This is needed to allow `osd-network-verifier` to properly adopt boilerplate, and is currently failing running rehearsals due to permissions issues https://github.com/openshift/release/pull/63612. ONV does not use the operator convention, but does use these two which are missing the fix: `golang-lint`, and `golang-codecov`.

This also updates `GOENV` to explicitly use `GOCACHE` that is the `/tmp/home` directory we set as `pwd` earlier.

[Sample Failure](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/63612/rehearse-63612-pull-ci-openshift-osd-network-verifier-main-build/1909721074057089024)

### DIFF
--- a/boilerplate/openshift/golang-codecov/standard.mk
+++ b/boilerplate/openshift/golang-codecov/standard.mk
@@ -9,10 +9,11 @@ GOFLAGS_MOD ?=
 # as $HOME is set to "/" by default.
 ifeq ($(HOME),/)
 export HOME=/tmp/home
+GOENV+=GOCACHE="${HOME}/.cache"
 endif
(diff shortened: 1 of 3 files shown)


dustman9000:
/lgtm
Verified changes locally on AVO

PR #520, opened 2025-04-15
Handle `GOCACHE` better in CI environments

joshbranham:
Our previous attempts to handle an unwritable `GOCACHE` weren't solving 100% of the cases, and leading to failures when I was attempting to update `osd-network-verifier`.

To ensure we always have a writeable `GOCACHE`, this PR modifies (and further extends) checks for `$HOME/.cache` not being writeable, and if so, creates a temporary directory and updates `GOCACHE` as such.

We then use this `GOCACHE` value via `GOENV` when we call `go` or `golangci-lint`.

Since Make is executing directives at _parse time_, just setting `GOCACHE=foo` early won't work on it's own, and callers have to explicitly check and use a custom tmp dir. I have done this for all the existing conventions, so users won't need to modify their per-repo `Makefile`s.

This will ensure that locally your `go env GOCACHE` value is still used, and in CI we always have a writeable tmp dir.

A sample PR can be found here utilizing these changes, showing passing CI. https://github.com/openshift/osd-network-verifier/pull/305

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ This work was inspired by, and partially cribbed from,
     - [Update](#update)
     - [Multiple Updates](#multiple-updates)
   - [Contributing](#contributing)
-    - [Environment setup](#environment-setup)
+    - [Testing Boilerplate Locally](#testing-boilerplate-locally)
     - [Tests](#tests)
     - [Build Images](#build-images)
       - [Making CI Efficient](#making-ci-efficient)
@@ -339,11 +339,10 @@ In your fork of this repository (not a consuming repository):
     - `LATEST_IMAGE_TAG`: The tag for the most recent build image
       produced by boilerplate.
 
-### Environment setup
+### Testing Boilerplate Locally
 To test your changes, you can use the `BOILERPLATE_GIT_REPO` environment
 variable and set it to your local clone in order to override the version of
 boilerplate used (Example: `export BOILERPLATE_GIT_REPO=~/git/boilerplate`).
-
 
 Default `update` behaviour consists of cloning the git repo, so ensure you have
 your changes locally committed for your testing.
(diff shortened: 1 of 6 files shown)


boilerplate/_lib/boilerplate.mk
```diff
@@ -1,3 +1,11 @@
+# When running in a CI environment, we usually can't write to the default go cache of
+# ${HOME}/.cache/go-build. This file is the first chunk of Makefile loaded from a repo's `generated-includes.mk`,
+# therefore we compute this upfront and then check it in the various conventions boilerplate supports.
+ifdef PROW_JOB_ID
```
dustman9000, line 4: Can we just check if `/tmp` is writable?
```suggestion
if [ -w "/tmp" ]; then
```

joshbranham: Are you suggesting that if `/tmp` is writeable we should use it for the cache 100% of the time?

The intent with this change is to:

- Let existing `GOCACHE` behavior work in local dev, so users existing `$HOME/.cache/go-build` can be used, versus all `make` targets re-pulling dependencies every time
- In CI environments, if `go env GOCACHE` returns `off`, that means the default directory `$HOME/.cache/go-build` is not writeable, so this was just a way to always ensure that is writeable up front without checking it in all the various conventions

I played with a few iterations but the best I could come up with was just keying off CI vs local to handle both (ignoring a users local GOCACHE value being broken). Open to other options, but I don't think just using `/tmp` if writeable solves it all.

joshbranham: @dustman9000 I have made some changes, which also allowed the existing tests to pass (which are proving very difficult to use as they fail to even run locally).

PR #523, opened 2025-04-15
OSD-29557: Remove the `GOLANGCI_OPTIONAL_CONFIG` option, which is unused

joshbranham:
In https://github.com/openshift/boilerplate/pull/92 we added an option to be able to pass a custom golangci-lint configuration on a per repo basis. I believe the intent is to have it be _additive_, in that it would run a second `golangci-lint` invocation with your custom config. This would only work if you added a linter, as removing one in this file still meant the original invocation would have failed with the boilerplate-based config.

Additionally, the gate that was designed to prevent `golangci-lint` from running a second time, unless `GOLANGCI_OPTIONAL_CONFIG` is set, appears to not be working, so we are currently running linting twice on all our repos.

Lastly, after doing a quick search of GitHub, it appears we only have to repos with the configuration:
- https://github.com/openshift/osd-metrics-exporter/ configures it but the file was previously deleted, so no-ops
- https://github.com/openshift/osd-cluster-ready configures it, but the file just enables `gosec` that is part of our base config

For the above reasons, this PR proposes removing this feature to simplify linting versus fixing bugs with the existing, unused process. If approved, I will then go to the above two repos and remove the config.

### DIFF
--- a/boilerplate/openshift/golang-lint/README.md
+++ b/boilerplate/openshift/golang-lint/README.md
@@ -13,5 +13,3 @@
 - ensures the proper version of `golangci-lint` is installed, and
 - runs it against
 - a `golangci.yml` config.
-- a `GOLANGCI_OPTIONAL_CONFIG` config if it is defined and file exists
-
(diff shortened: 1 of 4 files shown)


rafael-azevedo:
+1 for removing this feature. unnecessary overhead not being used.

PR #527, opened 2025-04-17
SDCICD-1472 skip must-gather and destroy cluster steps in operator e2e suite

ritmun:
Aimed at making e2e jobs in PD pipelines  faster
- skip-destroy-cluster will skip uninstall logic in test. clusters will still be destroyed by ocm at their expiration 30 min after test
- must gather is not saved anywhere, so is useless in these jobs 
- changes will take effect only when boilerplate is updated on operator repos

### DIFF
--- a/boilerplate/openshift/golang-osd-operator-osde2e/test-harness-template.yml
+++ b/boilerplate/openshift/golang-osd-operator-osde2e/test-harness-template.yml
@@ -46,6 +46,8 @@ objects:
               args:
                 - test
                 - --only-health-check-nodes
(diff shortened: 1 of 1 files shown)


jbpratt (approved):
/lgtm

joshbranham:
> skip-destroy-cluster will skip uninstall logic in test. clusters will still be destroyed by ocm at their expiration 30 min after test

Is this how other jobs handle this? It seems a bit messy to not clean up infra after a CI run. Is there some delay in e2e such that we need to keep it under a certain duration?

PR #542, opened 2025-05-16
pipelines/bundle-builder: add missing tasks

jbpratt:
add sast-shell-check and sast-unicode-check, they are required


joshbranham (commented):
/lgtm

PR #565, opened 2025-07-01
SDCICD-1593 login to quay repo before push

ritmun:
login to quay repo before push - login statement missing from simplified makefile

### DIFF
--- a/boilerplate/openshift/golang-osd-e2e/standard.mk
+++ b/boilerplate/openshift/golang-osd-e2e/standard.mk
@@ -66,7 +66,7 @@ e2e-binary-build:
 
 # push e2e image tagged as latest and as repo commit hash
 .PHONY: e2e-image-build-push
-e2e-image-build-push:
+e2e-image-build-push: container-engine-login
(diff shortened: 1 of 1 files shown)


boilerplate/openshift/golang-osd-e2e/standard.mk
```diff
@@ -66,7 +66,7 @@ e2e-binary-build:
 
 # push e2e image tagged as latest and as repo commit hash
 .PHONY: e2e-image-build-push
-e2e-image-build-push:
+e2e-image-build-push: container-engine-login
```
joshbranham, line 69: The previous version of this make target used a script and didn't login first https://github.com/openshift/boilerplate/blob/1b42d876c286bfbbfef9efcb7b01ce331f7a9afa/boilerplate/openshift/golang-osd-operator-osde2e/standard.mk#L78

Can you explain a bit why this changed and why we need to run this make target `container-engine-login` now?

joshbranham (commented):
/lgtm

PR #566, opened 2025-07-02
Cleanup make targets and fix LICENSE copy

joshbranham:
In https://github.com/openshift/boilerplate/commit/07bc19ce54b73298991228c6b81e484c653780b7 we added a `COPY` for the `LICENSE` file. I have noticed I fail to build locally because of this. But, that begs the question: why does it work in Konflux?

I cleaned up the make targets because I see no reference to `build-image-shallow` in `openshift/release` or our Konflux pipelines.

@jbpratt for confirmation on these changes


Makefile
```diff
@@ -35,11 +35,7 @@ subscriber-report: ## Discover onboarding and prow status of subscribed consumer
 
 .PHONY: build-image-deep
 build-image-deep: ## Builds the image from scratch, like appsre does. May require ALLOW_DIRTY_CHECKOUT=true if testing
-	cd config; $(CONTAINER_ENGINE) build -t $(IMG):latest -f Dockerfile .
-
-.PHONY: build-image-shallow
-build-image-shallow: ## Builds the image starting from a recent release, like prow does. May require ALLOW_DIRTY_CHECKOUT=true if testing
-	cd config; $(CONTAINER_ENGINE) build -t $(IMG):latest .
+	$(CONTAINER_ENGINE) build -t $(IMG):latest -f config/Dockerfile .
```
joshbranham, line 38: Set the buildroot to the root of the checkout where the `LICENSE` is

PR #575, opened 2025-07-18
SREP-1170: Create Tekton Pipeline to run `pr-check` in Konflux

joshbranham:
This does two things:

- Add a new pipeline that we can configure in Konflux to run an integration test (`make pr-check`) as we move this from prow to Konflux. This will be configured to run after an image is built for a PR.
- Remove all the hacks in the tests that were designed around shortcomings in the way `prow` ran the job before.

I ran this locally with this PipelineRun config

```
kind: PipelineRun
metadata:
  name: test-run
spec:
  pipelineRef:
    name: boilerplate-master-pr-check
  workspaces:
    - name: shared-workspace
      volumeClaimTemplate:
        spec:
          accessModes:
            - ReadWriteOnce
          resources:
            requests:
              storage: 1Gi
```

Two follow ups:

- Delete the prow config so it doesn't run `pr-check`
- Add the integration test config for Konflux

### DIFF
--- a/.tekton/boilerplate-master-pr-check.yaml
+++ b/.tekton/boilerplate-master-pr-check.yaml
@@ -1,0 +1,83 @@
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: boilerplate-master-pr-check
+  labels:
+    build.appstudio.redhat.com/pipeline: boilerplate-master-pr-check
+spec:
+  params:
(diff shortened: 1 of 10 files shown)


jbpratt:
/lgtm
/hold

holding so you can merge when you want

joshbranham:
/override Red Hat Konflux / boilerplate-master-pr-check / image

PR #576, opened 2025-07-18
Drop no longer needed .ci-operator.yaml config

joshbranham:
We are no longer doing presubmits via prow so drop this config


jbpratt (approved):
/lgtm

PR #580, opened 2025-07-19
SREP-1170: Ensure `make pr-check` works on macOS and in Konflux

joshbranham:
The fixes below were made with the assistance of Cursor:

- `find` support for GNU and BSD
- `mktemp` support for GNU and BSD
- Color escape sequence fixes so terminal output shows the expected colors
- Ensure `tag-check.sh` works even when there is no `master` branch ref in the checkout, as is the case in Konflux.

### DIFF
--- a/.tekton/boilerplate-master-pr-check.yaml
+++ b/.tekton/boilerplate-master-pr-check.yaml
@@ -54,7 +54,7 @@ spec:
         - name: fetchTags
           value: "true"
         - name: depth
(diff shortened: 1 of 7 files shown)


joshbranham:
https://gitlab.cee.redhat.com/releng/konflux-release-data/-/merge_requests/8636 contains the final piece to wire this all up, which I have modified and taken off the Konflux cluster directly.

PR #583, opened 2025-07-22
Drop defunct tag-check.sh

joshbranham:
We don't really rely on this `tag-check.sh` script anymore, as we tag a commit on `master` after merge for Konflux to build a release from.


README.md
```diff
@@ -414,12 +414,6 @@ to create boilerplate:image-v3.0.0
     FROM registry.ci.openshift.org/openshift/boilerplate:image-v3.0.0
     ```
 
-#### Picking Up (Security) Fixes
```
joshbranham, line 417: The process is just to follow the Konflux release process.

PR #588, opened 2025-07-28
Ensure consumers FROM statements in Dockerfiles are using ubi9

joshbranham:
In https://github.com/openshift/boilerplate/commit/6b3d4fabac7ea7744a674957ab0e80dad4d41ddd we updated this logic to iterate `FROM` directives in a consuming projects `Dockerfile` and update the UBI tag to the latest for a given major version (7,8,9 etc).

Now that we have update boilerplate to UBI9, in hindsight, this logic existed to _force_ consumers to use the same UBI version that boilerplate expects. If they don't, compiled binaries will fail to run.

This reverts the above commit, while also adding some context as to how this is load bearing for the future.

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/update
+++ b/boilerplate/openshift/golang-osd-operator/update
@@ -92,10 +92,12 @@ for file in $DOCKERFILES; do
     ${SED?} -i "1s,.*,FROM $IMAGE_PULL_PATH AS builder," $file
   fi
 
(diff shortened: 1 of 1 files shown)


joshbranham:
The image is failing to build due to possibly rate limiting by GitHub. We don't need to build the image here anyways, we are updating minor logic that won't affect the image. I will override once approved.

PR #605, opened 2025-08-28
SDCICD-1643: optimize presubmits

jbpratt:
only run if the change calls for it

Signed-off-by: Brady Pratt <bpratt@redhat.com>


joshbranham (commented):
Nice!

/lgtm

PR #716, opened 2026-04-22
SREP-4482: Update golangci-lint configuration with enhanced linters

devppratik:
Based on the agreed golden standard for golangci.yml which can be used by all the agents & developers, enhance the golangci-lint configuration to include a more comprehensive set of linters organized by priority (Critical, High, Medium, Optional) with appropriate settings for error handling, security, and code quality checks.

Related to [SREP-4482](https://redhat.atlassian.net/browse/SREP-4482)

[SREP-4482]: https://redhat.atlassian.net/browse/SREP-4482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/golangci.yml
+++ b/boilerplate/openshift/golang-osd-operator/golangci.yml
@@ -1,39 +1,71 @@
 version: "2"
-run:
-  concurrency: 10
+
 linters:
-  default: none
   enable:
+    # Critical: Error Handling & Security
     - errcheck
-    - gosec
     - govet
-    - ineffassign
-    - misspell
     - staticcheck
+    - gosec
+    - bodyclose
+    - sqlclosecheck
+    - contextcheck
+    - noctx
+
+    # High: Error Prevention
+    - errorlint
+    - nilerr
+    - nilnil
+    - revive
+
+    # Medium: Code Quality
+    - ineffassign
+    - unconvert
+    - unparam
     - unused
+    - misspell
+
+    # Optional: Maintainability
+    - prealloc
+    - nolintlint
+    - gocyclo
+    - dupl
+    - exhaustive
+    - makezero
+    - containedctx
+
   settings:
+    revive:
+      rules:
+        - name: package-comments
+          disabled: true
+
+    errcheck:
+      check-type-assertions: true
+      check-blank: false
+      disable-default-exclusions: true
+
+    gocyclo:
+      min-complexity: 15
+
+    errorlint:
+      errorf: true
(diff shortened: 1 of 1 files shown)


boilerplate/openshift/golang-osd-operator/golangci.yml
```diff
@@ -1,39 +1,71 @@
...
-      - common-false-positives
-      - legacy
-      - std-error-handling
-    paths:
-      - third_party/
-      - builtin/
-      - examples/
+
+run:
+  timeout: 5m
```
joshbranham, line 62: What is the motivation for removing the `concurrency` and setting a 5m timeout?

devppratik: From https://golangci-lint.run/docs/configuration/file/#run-configuration
In golangci-lint v2, the default concurrency is 0, which automatically matches the Linux container CPU quota and falls back to the number of logical CPUs. Removing the explicit concurrency: 10 setting allows golangci-lint to auto-tune based on available resources, which is the recommended approach.


I have added the timeout setting to prevents indefinite run. It seemed to work without any issues in the local. Can adjust as needed

joshbranham: Gotcha, so this is a global config meaning each operator will use it. If we run into a project taking too long, we will need to tune this timeout.

Can you test this config on some other projects, like `osdctl` etc to just get a bit more data to prove the 5m is enough? I have seen timeouts in CI for linting due to resource pressure in the CI clusters.

geowa4: While I personally prefer having the CI or surrounding framework (like prek or pre-commit) handle the per-task timeout, 5 minutes seems like a reasonable starting point to build off of, and testing on others like osdctl seems like a reasonable request to get a multi-project baseline. 

Other thought: we should fail builds if this times out so we can get that feedback and make the correction.

devppratik: @joshbranham Yes, this is more of a global config that we add. We can look. to have this fine tune later per project if required. I did check with osdctl locally and it seems to be running without issues.

geowa4: Ran this on a couple other projects and measured time in seconds so 5m should have plenty of buffer. LGTM

PR #720, opened 2026-04-29
SREP-4484: Enable Codecov Enforcement

devppratik:
As a part of [SREP-4484](https://redhat.atlassian.net/browse/SREP-4484), Enable and configure codecov status checks with coverage thresholds

[SREP-4484]: https://redhat.atlassian.net/browse/SREP-4484?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/.codecov.yml
+++ b/boilerplate/openshift/golang-osd-operator/.codecov.yml
@@ -8,8 +8,14 @@ coverage:
   range: "20...100"
 
   status:
-    project: no
-    patch: no
+    project:
+      default:
+        target: 50%
+        threshold: 1%
+    patch:
+      default:
+        target: 50%
+        threshold: 1%
     changes: no
 
 parsers:
--- a/boilerplate/openshift/golang-osd-operator/.codecov.yml
+++ b/boilerplate/openshift/golang-osd-operator/.codecov.yml
@@ -14,8 +14,8 @@ coverage:
         threshold: 1%
     patch:
       default:
-        target: 60%
-        threshold: 5%
+        target: 50%
+        threshold: 1%
     changes: no
 
 parsers:
--- a/boilerplate/openshift/golang-osd-operator/.codecov.yml
+++ b/boilerplate/openshift/golang-osd-operator/.codecov.yml
@@ -10,12 +10,12 @@ coverage:
   status:
     project:
       default:
-        target: 50%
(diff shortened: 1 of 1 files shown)


boilerplate/openshift/golang-osd-operator/.codecov.yml
```diff
@@ -8,8 +8,14 @@ coverage:
...
-    project: no
-    patch: no
+    project:
+      default:
+        target: 50%
+        threshold: 1%
+    patch:
+      default:
+        target: 50%
+        threshold: 1%
```
geowa4: Is there a way to ensure the code coverage doesn't decrease instead of using a fixed number? With coding agents, we want to see the percentage stay the same or go up. This mitigates the risk that unattended coding agents go wild generating untested and untestable code.

devppratik: Patch coverage = coverage of new/changed lines only. Ensures any new code added by agents is tested. We can increase the target % to ensure new code is tested

geowa4: That seems reasonable, thanks! Do we need the project default then? If a project is at 20% does, the build fail because the coverage is too low overall? I don't want to punish a project for low test coverage _today_. I want to punish them for not improving or holding constant.

joshbranham: I agree with George, we only want to encourage test coverage for new code. However, I don't want unit tests that test the compiler, so we need to come up with a pattern here that doesn't give us unnecessary test bloat. One solution could just be making codecov optional, or, overriding in cases where a maintainer disagrees that more tests are needed.

devppratik: @geowa4 The current project default is at 50%. I can reduce it to about 40%. I did a audit of repos and it seems almost of our repos are more than 40% at the test coverage

> One solution could just be making codecov optional, or, overriding in cases where a maintainer disagrees that more tests are needed.

@joshbranham Agreed. While this is the base common config, repos should be able to update the values themselves

geowa4: Works for me. Ship it.

PR #731, opened 2026-05-08
Re-enable std-error-handling exclusion for golang-osd-operator lint

tnierman:
Recent updates to the `golang-osd-operator`'s golangci config removed the std-error-handling exclusion preset, causing errcheck to flag unchecked return values from standard library functions like `fmt.Printf` and `fmt.Println`. 

The golang-lint convention already includes this preset. Also removes disable-default-exclusions which was redundant with the explicit preset.

🤖 Generated with Claude Code (https://claude.ai/claude-code)


joshbranham (commented):
/lgtm
/approve

PR #734, opened 2026-05-11
Fix incremental linting for enhanced golangci config

abyrne55:
The enhanced golangci-lint config added in d83e5ee uses `new: true` to only lint new code, but this silently falls back to linting everything in CI shallow clones where the required git history is unavailable (see golangci/golangci-lint#4180). This breaks `ci/prow/lint` across all consumer repos on boilerplate update.

Both `new: true` and `new-from-rev` in the YAML config have this same problem — they rely on git history that shallow clones don't provide, and golangci-lint silently falls back to linting the entire codebase.

**Fix:** Pass `--new-from-rev` via `standard.mk`'s `go-check` target using a ref that's guaranteed to exist:
- In Prow CI, use `PULL_BASE_SHA` (always present and valid, even in shallow clones)
- Locally, fall back to `origin/HEAD` via `git symbolic-ref`

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/golangci.yml
+++ b/boilerplate/openshift/golang-osd-operator/golangci.yml
@@ -62,11 +62,9 @@ linters:
 
 run:
   timeout: 5m
-  # Only check new code, not existing issues
-  # This uses git to compare against the base branch (typically master/main)
-  new: true
-  # Alternatively, you can specify a specific revision:
-  # new-from-rev: origin/master
+  # Incremental linting (new-from-rev) is passed via the Makefile's
+  # go-check target. In CI it uses PULL_BASE_SHA (guaranteed to exist
+  # even in shallow clones); locally it falls back to origin/HEAD.
 
 formatters:
   enable:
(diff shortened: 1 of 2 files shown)


abyrne55:
Tested both code paths in standard.mk's go-check target against aws-vpce-operator (a consumer repo that was failing ci/prow/lint):

1. **Built the consumer repo with this fix** — ran `make boilerplate-update` with `BOILERPLATE_GIT_REPO` pointed at this branch, confirmed the `standard.mk` and `golangci.yml` changes landed correctly.

2. **Local path** (`origin/HEAD` fallback) — ran `make go-check` without `PULL_BASE_SHA`. The Makefile resolved `LINT_NEW_FROM_REV` to `origin/main` via `git symbolic-ref`, passed `--new-from-rev=origin/main` to golangci-lint. Result: **0 issues**.

3. **CI path** (`PULL_BASE_SHA`) — ran `PULL_BASE_SHA=f5ede65 make go-check` to simulate Prow. The Makefile used the env var directly, passed `--new-from-rev=f5ede65`. Result: **0 issues**.

Both paths previously produced 200+ lint failures because `new: true` silently fell back to linting the entire codebase.

PR #742, opened 2026-05-26
chore(Update pre-commit-yaml)

devppratik:
Updating pre-commit yaml based on feedback received in https://github.com/openshift/splunk-forwarder-operator/pull/430 to rollout to all operators

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/docs/pre-commit.md
+++ b/boilerplate/openshift/golang-osd-operator/docs/pre-commit.md
@@ -1,0 +1,123 @@
+# Pre-Commit Hooks Setup Guide
+
+## Installation
+
+### Recommended: Using uv
+
+[uv](https://github.com/astral-sh/uv) is recommended for Python dependency management. It provides dependency locking with package hashes (supply-chain protection), virtual environment management, and is 10-100x faster than pip.
+
+**Install uv:**
+```bash
+# macOS/Linux
+curl -LsSf https://astral.sh/uv/install.sh | sh
+
+# Windows
+powershell -c "irm https://astral.sh/uv/install.ps1 | iex"
+
+# Via pip
+pip install uv
+```
+
+**First-time setup:**
+```bash
+uv init --bare                    # creates pyproject.toml
+uv add --dev pre-commit==4.6.0    # adds dependency, generates uv.lock
+source .venv/bin/activate         # macOS/Linux (.venv\Scripts\activate on Windows)
+pre-commit install
+```
+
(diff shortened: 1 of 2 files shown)


boilerplate/openshift/golang-osd-operator/docs/pre-commit.md
```diff
@@ -0,0 +1,123 @@
...
+powershell -c "irm https://astral.sh/uv/install.ps1 | iex"
+
+# Via pip
+pip install uv
+```
+
+**First-time setup:**
+```bash
+uv init --bare                    # creates pyproject.toml
+uv add --dev pre-commit==4.6.0    # adds dependency, generates uv.lock
```
geowa4, line 24: Why not use uvx?

boilerplate/openshift/golang-osd-operator/docs/pre-commit.md
```diff
@@ -0,0 +1,123 @@
+# Pre-Commit Hooks Setup Guide
+
+## Installation
```
geowa4, line 3: The length of these installation instructions for pre-commit are one reason why I prefer prek. This isn't a blocker but something to consider.

boilerplate/openshift/golang-osd-operator/docs/pre-commit.md
devnulljason: [`uvx` docs here](https://docs.astral.sh/uv/concepts/tools/). It may be preferable, but we'll want to see if we can still pin to the tool's hash like you can with the version.

Along those same lines, I would prefer if we could control the version of the tool through the `pyproject.toml` file instead of just having it in user instructions. Either by copying an entire `pyproject.toml` to the repo's base directory if it doesn't already exist or adding it to the file if it does.

EDIT: It appears referencing the version number of the tool is the best way to use it with `uvx`, so instead of these two commands you would just do:
```bash
$ uvx pre-commit@4.6.0 <command>
```

boilerplate/openshift/golang-osd-operator/docs/pre-commit.md
devppratik: We can start here for now. I am exploring prek and would update if those become better working with Claude

boilerplate/openshift/golang-osd-operator/docs/pre-commit.md
devppratik: As mentioned in the previous comment, I am still looking at prek as an alternative. For now the uv command would work. If prek becomes more easier to handle, it would replace the pre-commit-yaml

PR #754, opened 2026-06-09
Remove srep-infra-cicd from golang-osd-operator update script [SDCICD-1761]

christophermancini:
The srep-infra-cicd team migration is complete. Operator teams should now own their Konflux setup, so we no longer automatically add srep-infra-cicd to .tekton/OWNERS files during boilerplate updates.

This change updates the template code so future bp-update runs will intelligently remove our team from the OWNERS file if it exists.

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/update
+++ b/boilerplate/openshift/golang-osd-operator/update
@@ -18,15 +18,25 @@ cp ${HERE}/.codecov.yml $REPO_ROOT
 echo "Copying OWNERS_ALIASES to your repository root."
 cp -L ${HERE}/OWNERS_ALIASES $REPO_ROOT
 
-# Add CICD owners to .tekton if exists
-if [[ -d "${REPO_ROOT}/.tekton/" ]]; then
-	echo "Adding Konflux subdirectory OWNERS file to .tekton/"
-	cat >"${REPO_ROOT}/.tekton/OWNERS" <<EOF
-reviewers:
-- srep-infra-cicd
-approvers:
-- srep-infra-cicd
-EOF
+# Clean up srep-infra-cicd from .tekton/OWNERS if it exists
+if [[ -f "${REPO_ROOT}/.tekton/OWNERS" ]]; then
+	echo "Checking .tekton/OWNERS for srep-infra-cicd cleanup..."
+
+	# Check if file contains srep-infra-cicd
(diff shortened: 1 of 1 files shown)


ritmun (commented):
reviewing

ritmun (changes requested):

ritmun (commented):
The current code has a complex piped `grep` chain to count other owners. Using the splunk-forwarder-operator as an example (where only `srep-infra-cicd` is present → file should be deleted), the logic can be simplified.

The key issue with the current code: the `other_owners` counting with multiple `grep -v` pipes is fragile and hard to read. A cleaner approach would be to:
1. Remove `srep-infra-cicd` lines first
2. Then check if any meaningful content remains

boilerplate/openshift/golang-osd-operator/update
```diff
@@ -18,15 +18,25 @@ cp ${HERE}/.codecov.yml $REPO_ROOT
...
+		if [[ "$other_owners" -eq 0 ]]; then
+			# Only srep-infra-cicd present, delete the file
+			echo "Removing .tekton/OWNERS (only contained srep-infra-cicd)"
+			rm "${REPO_ROOT}/.tekton/OWNERS"
+		else
+			# Other owners present, just remove srep-infra-cicd lines
+			echo "Removing srep-infra-cicd from .tekton/OWNERS (keeping other owners)"
+			sed -i '/srep-infra-cicd/d' "${REPO_ROOT}/.tekton/OWNERS"
+		fi
+	fi
```
ritmun, line 31:
```suggestion
# Clean up srep-infra-cicd from .tekton/OWNERS if it exists
TEKTON_OWNERS="${REPO_ROOT}/.tekton/OWNERS"
if [[ -f "${TEKTON_OWNERS}" ]] && grep -q "srep-infra-cicd" "${TEKTON_OWNERS}"; then
	echo "Removing srep-infra-cicd from .tekton/OWNERS..."
	${SED?} -i '/srep-infra-cicd/d' "${TEKTON_OWNERS}"

	# If no owners remain (ignoring section headers, comments, and blank lines), remove the file
	if ! grep -qE "^\\s*-\\s+\\S" "${TEKTON_OWNERS}"; then
		echo "Removing .tekton/OWNERS (no owners remain after cleanup)"
		rm "${TEKTON_OWNERS}"
	fi
fi
```

christophermancini: Good idea! Accepting your suggestion.

PR #767, opened 2026-06-19
Update golangci-lint to v2.12.2 for Go 1.26 support

rpodishe:
Update golangci-lint to v2.12.2 for Go 1.26 support


joshbranham:
Fixed by https://github.com/openshift/boilerplate/pull/765

PR #804, opened 2026-07-17
Fix the default branch detection bug as CI doesn't fully clone the repo

tkong-redhat:
In the freeze-check context, the boilerplate repo is fetched via git fetch origin <commit> --tags (not a full clone), so `refs/remotes/origin/HEAD` never gets created. Then the both `git symbolic-ref` call will fall back to master which failed the CI validation check.
Update the default branch detection from existing pipeline file align with other parameters.


tkong-redhat:
/override "Red Hat Konflux / boilerplate-enterprise-contract / image"

PR #807, opened 2026-07-17
Harden gangway-bridge template

dustman9000:
Address review feedback on the gangway-bridge template:

1. **Dynamic activeDeadlineSeconds**: Use ${{TIMEOUT}} instead of hardcoded 7200 so the Job deadline matches the configured timeout
2. **Validate numeric parameters**: Check TIMEOUT and POLL_INTERVAL are numeric before bash arithmetic to prevent injection
3. **Read-only root filesystem**: Add readOnlyRootFilesystem: true since the polling Job doesn't need writes


joshbranham (commented):
/lgtm

PR #817, opened 2026-07-22
Fix pipeline ref pinning to use exported BOILERPLATE_COMMIT

dustman9000:
Fixes the pipeline ref pinning introduced in #816. The previous approach used git -C HERE rev-parse HEAD in the convention update script, but HERE points to a directory inside the consuming repo (after rsync), so git resolved to the consuming repo HEAD instead of the boilerplate clone.

Fix: export BOILERPLATE_COMMIT from the main boilerplate/update script (same pattern as LATEST_IMAGE_TAG) so convention update scripts can reference it directly.

Tested locally: ran BOILERPLATE_GIT_REPO=~/src/boilerplate make boilerplate-update and verified the generated .tekton pipeline revision matches last-boilerplate-commit.


geowa4 (approved):
/lgtm

dustman9000:
/override "Red Hat Konflux / boilerplate-enterprise-contract / image"

dustman9000:
/override "Red Hat Konflux / boilerplate-enterprise-contract / image"

PR #842, opened 2026-08-13
Resolve build image tag to newest published, not newest git tag

AlexSmithGH:
Publishing an `image-v*` tag does not publish an image; it only lets you start the Konflux release that builds one. Until that release completes, the tag exists in git with nothing behind it.

`update` resolved LATEST_IMAGE_TAG with `git describe --match image-v*`, so during that window consumers recorded a tag with no image and wrote it into build/Dockerfile's FROM, .ci-operator.yaml's build root, and IMAGE_PULL_PATH. Their builds then failed on a nonexistent reference, and freeze-check pinned them there until someone updated again. This happened for real between image-v8.4.2 (tagged, never released) and image-v8.4.3.

Walk tags newest-first instead and take the first one actually present in the registry, so an update mid-release yields the previous working image.

Only tags reachable from the checked-out commit are considered. `git describe` was ancestry-scoped and a bare `git tag --list` is not; without `--merged` freeze-check would resolve tags published after the commit it replays and fail for every consumer.

The skopeo call is inlined rather than reusing _lib/common.sh's image_exists_in_repo(): that library derives LATEST_IMAGE_TAG from the consumer's existing _data/backing-image-tag, which would short-circuit the lookup and pin consumers to their old tag permanently. Sourcing it would also pull linguist-generated code into a script .gitattributes deliberately keeps visible in PR diffs.

Only "manifest unknown" means unpublished. Any other registry error is fatal, since treating an outage or auth failure as "not published" would silently walk back to an arbitrarily old image. The walk is depth-bounded so an outage fails fast.

test/lib.sh set LATEST_IMAGE_TAG without exporting it, so the driver never saw it; the whole suite would now hit the live registry. Export it to keep tests offline and deterministic. The assertion it originally fed was removed in 26fbdba.

## Summary by CodeRabbit

* **Documentation**
  * Clarified how the latest image tag is selected and how to pin a specific image version.
  * Added release guidance explaining image publication timing and fallback behavior.

* **Tests**
  * Added deterministic coverage for image-tag selection, overrides, unpublished and unreachable tags, missing images, and registry errors.
  * Improved test configuration to support pinned image tags and registry lookup scenarios.

### DIFF
--- a/boilerplate/update
+++ b/boilerplate/update
@@ -161,8 +161,64 @@ if [ ! -f "$CONFIG_FILE" ]; then
 fi
 
 # The most recent build image tag. Export this for individual `update` scripts.
+#
+# A git tag alone doesn't mean the image is consumable: publishing the tag only
+# starts the Konflux release that builds and pushes the image. Picking the
+# newest tag unconditionally means that, for as long as that release takes,
+# consumers get a tag with no image behind it -- which breaks their
+# build/Dockerfile FROM, their .ci-operator.yaml build root, and
+# `container-make`. So walk tags newest-first and take the first one that
+# actually exists in the registry.
+#
+# Only tags reachable from the checked-out commit are considered. This keeps
+# `freeze-check` reproducible: it re-runs this driver against the commit
+# recorded in _data/last-boilerplate-commit, and must not resolve a tag that
+# was published after that commit.
+#
+# NOTE: This is deliberately inlined rather than sourced from _lib/common.sh.
+# That library derives LATEST_IMAGE_TAG from the consumer's existing
+# _data/backing-image-tag, which would defeat the resolution below; and this
+# script is intentionally kept auditable in PR diffs via .gitattributes.
 if [[ -z "$LATEST_IMAGE_TAG" ]]; then
-    export LATEST_IMAGE_TAG=$(cd $BP_CLONE; git describe --tags --abbrev=0 --match image-v*)
+    # Bound the walk so a registry outage fails fast instead of trawling
+    # through every historical tag.
+    IMAGE_TAG_SEARCH_DEPTH=${IMAGE_TAG_SEARCH_DEPTH:-10}
+    if ! command -v skopeo &>/dev/null; then
+        echo "Failed to find the skopeo binary, needed to verify the build image tag." >&2
+        echo "If you are on Mac: brew install skopeo." >&2
+        exit 1
+    fi
+    candidates=$(cd $BP_CLONE; git tag --merged HEAD --sort=-v:refname --list 'image-v*' | head -n $IMAGE_TAG_SEARCH_DEPTH)
+    for tag in $candidates; do
+        image_uri="quay.io/redhat-services-prod/openshift/boilerplate:${tag}"
+        skopeo_stderr=$(mktemp)
+        if skopeo inspect --override-os linux --override-arch amd64 "docker://${image_uri}" >/dev/null 2>$skopeo_stderr; then
+            rm -f $skopeo_stderr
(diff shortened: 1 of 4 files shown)


boilerplate/update
```diff
@@ -161,8 +161,64 @@ if [ ! -f "$CONFIG_FILE" ]; then
...
+#
+# Only tags reachable from the checked-out commit are considered. This keeps
+# `freeze-check` reproducible: it re-runs this driver against the commit
+# recorded in _data/last-boilerplate-commit, and must not resolve a tag that
+# was published after that commit.
+#
+# NOTE: This is deliberately inlined rather than sourced from _lib/common.sh.
+# That library derives LATEST_IMAGE_TAG from the consumer's existing
+# _data/backing-image-tag, which would defeat the resolution below; and this
+# script is intentionally kept auditable in PR diffs via .gitattributes.
```
joshbranham, line 181: Can we just make this way simpler by having it check quay for existence of the latest tag it discovered, and if it 404s, don't update?

AlexSmithGH:
Good instinct, and it would be less code — but "don't update" can't be honored at the point we learn the tag is missing, and it also changes the outcome in a way I don't think we want.

**The blocking problem: by the time we check, we've already updated.**

The driver re-execs itself. Before that re-exec, it has already overwritten `boilerplate/update` and blown away and re-synced the `_lib` utilities:

```bash
rsync -a "${BP_CLONE}/boilerplate/update" "$0"     # update:115
for d in $(/bin/ls -d ${BP_CLONE}/boilerplate/_*); do
  rm -fr $CONVENTION_ROOT/${d##*/}                 # update:121
done
rsync -a -r --delete ${BP_CLONE}/boilerplate/_* $CONVENTION_ROOT   # update:123
exec "$0" "$BP_CLONE"                              # update:126
```

Tag resolution happens at `update:~180`, i.e. after all of that. So exiting there doesn't mean "don't update" — it means the consumer is left with new `update`/`_lib` but stale conventions and a stale `_data/backing-image-tag`, which is a worse state than either outcome. Making it a true no-op would mean hoisting the registry check above line 115, which means resolving tags before we've cloned the repo we resolve them from. That's a bigger restructure than the loop it replaces.

**The behavioral problem: it converts a broken build into a blocked one.**

`image-v8.4.2` was tagged and never released. Real tag-vs-registry state today:

```
image-v8.3.3  published
image-v8.4.0  published
image-v8.4.1  published
image-v8.4.2  MISSING     <- tagged, release never completed
image-v8.4.3  published
```

Under this PR, someone updating during that window silently gets `image-v8.4.1` and their build works. Under "404 → don't update," everyone updating in that window is hard-blocked until a human notices `8.4.2` never released — and note `8.4.2` sat missing from Aug 10 to Aug 12. Given the whole point is closing the release gap, trading "broken update" for "no updates possible" doesn't feel like a win.

**On the simplicity delta:** it's smaller than it looks. Both versions need the same skopeo invocation and the same `manifest unknown`-vs-real-error discrimination (an auth failure or network blip must not be read as "unpublished"). The only thing that goes away is the loop and `--merged`/`--sort=-v:refname`, maybe 8 lines. And `--merged` is load-bearing for a separate reason: `freeze-check` replays this driver at the commit in `_data/last-boilerplate-commit`, so it must not see tags published after that commit.

Happy to be talked into it if the blocking behavior is actually what you want — but I'd want to do it as the hoisted-check restructure rather than an exit partway through, so the "don't update" is real. Alternatively, if the concern is mainly about the loop walking too far, I can drop `IMAGE_TAG_SEARCH_DEPTH` from 10 to something tight like 3, so it only ever covers an in-flight release or two.

PR #866, opened 2026-08-31
feat: add skip-roxctl-scan for scratch PKO images

MitaliBhalla:
## Summary
- Add optional `skip-roxctl-scan` on `docker-build-oci-ta`, same pattern as `skip-preflight-cert-check`
- ACS `roxctl-scan` (replaced Clair on 2026-08-28) fails on FROM-scratch PKO/package images
- `skip-checks: true` is the wrong workaround: it disables clamav/SAST/RPM scans and Enterprise Contract then fails

## Test plan
- [ ] Default remains `false`; operator images still run ACS
- [ ] PKO PipelineRuns can set `skip-roxctl-scan: true` and still run clamav/SAST/RPM signature
- [ ] Follow-up in rbac-permissions-operator: https://github.com/openshift/rbac-permissions-operator/pull/419

Made with [Cursor](https://cursor.com)

### DIFF
--- a/pipelines/docker-build-oci-ta/pipeline.yaml
+++ b/pipelines/docker-build-oci-ta/pipeline.yaml
@@ -71,6 +71,10 @@ spec:
       default: "false"
       type: string
       description: Skip the preflight cert-check. This can required for scratch images. Valid values are false and true.
+    - name: skip-roxctl-scan
+      default: "false"
+      type: string
+      description: Skip ACS/roxctl-scan only (keep clamav, SAST, RPM signature, etc.). Use for FROM-scratch PKO/package images that have no OS for ACS to scan. Valid values are false and true.
     - name: enable-package-registry-proxy
       default: 'true'
       description: Use the package registry proxy when prefetching dependencies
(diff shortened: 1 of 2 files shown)


pipelines/docker-build-oci-ta/pipeline.yaml
```diff
@@ -71,6 +71,10 @@ spec:
       default: "false"
       type: string
       description: Skip the preflight cert-check. This can required for scratch images. Valid values are false and true.
+    - name: skip-roxctl-scan
+      default: "false"
+      type: string
+      description: Skip ACS/roxctl-scan only (keep clamav, SAST, RPM signature, etc.). Use for FROM-scratch PKO/package images that have no OS for ACS to scan. Valid values are false and true.
```
joshbranham, line 77: Shouldn't ACS/`roxctl` just gracefully handle when there is an inability to scan?

joshbranham: Looks like it's being solved with https://redhat.atlassian.net/browse/KONFLUX-15651

PR #871, opened 2026-09-02
Add shared CodeRabbit configuration

bmeng:
Add a managed coderabbit config to exclude the common pattern which are not required for code review

### DIFF
--- a/boilerplate/openshift/golang-osd-operator/.coderabbit.yaml
+++ b/boilerplate/openshift/golang-osd-operator/.coderabbit.yaml
@@ -1,0 +1,14 @@
+inheritance: true
+reviews:
+  path_filters:
+    # Exclude boilerplate changes from review
+    - "!boilerplate/**"
+
+    # Exclude build artifacts and dependencies
+    - "!build/**"
+    - "!.venv/**"
+    - "!vendor/**"
+
+    # Exclude test fixtures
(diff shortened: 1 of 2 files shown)


boilerplate/openshift/golang-osd-operator/.coderabbit.yaml
```diff
@@ -0,0 +1,14 @@
+inheritance: true
+reviews:
+  path_filters:
+    # Exclude boilerplate changes from review
+    - "!boilerplate/**"
```
joshbranham, line 5: This is covered in the upstream coderabbit config in `openshift`

boilerplate/openshift/golang-osd-operator/update
```diff
@@ -14,6 +14,10 @@ source $CONVENTION_ROOT/_lib/common.sh
 echo "Copying .codecov.yml to your repository root."
 cp ${HERE}/.codecov.yml $REPO_ROOT
 
+# Add CodeRabbit configuration
+echo "Copying .coderabbit.yaml to your repository root."
+cp ${HERE}/.coderabbit.yaml $REPO_ROOT
```
joshbranham, line 19: What happens if a repo has a customized config, won't this wipe over it? We should try and not break per-repo customization.

PR #874, opened 2026-09-04
[fix] Preserve custom CodeRabbit configuration

bmeng:
Make the one-time sync for coderabbit.yaml to all the conventions

Simplify the logic to copy the file only if it is not present.


boilerplate/openshift/golang-osd-operator/update
```diff
@@ -14,9 +14,50 @@ source $CONVENTION_ROOT/_lib/common.sh
...
 
-# Add CodeRabbit configuration
-echo "Copying .coderabbit.yaml to your repository root."
-cp ${HERE}/.coderabbit.yaml $REPO_ROOT
+# Ensure the boilerplate path filter exists while preserving repo customizations.
+CODERABBIT_CONFIG="${REPO_ROOT}/.coderabbit.yaml"
+if [[ ! -f "${CODERABBIT_CONFIG}" ]]; then
+  echo "Copying .coderabbit.yaml to your repository root."
+  cp "${HERE}/.coderabbit.yaml" "${CODERABBIT_CONFIG}"
+elif grep -Fq '!boilerplate/**' "${CODERABBIT_CONFIG}"; then
```
joshbranham: for simplicity sake, would it make more sense to sync a .coderabbit.yaml if none exists (from a base template) and then ignore if it exists? Then folks can tune moving forward. Also it would be good to add this to other conventions in BP

PR #876, opened 2026-09-09
Update Go to 1.26.8 in builder image

anispate:
Replace `go-toolset` RPM installation (which currently provides Go 1.26.5 from UBI9 repos) with a direct installation of Go 1.26.8 from golang.org. This pins the Go version explicitly and ensures the builder provides Go 1.26.8.

`GOTOOLCHAIN=auto` is retained so consuming repositories can still override via their `go.mod` toolchain directive.

Part of https://redhat.atlassian.net/browse/ROSAENG-66308 and https://redhat.atlassian.net/browse/ROSAENG-66306


joshbranham:
We install go from `go-toolset` for FIPS compatibility, primarily.

PR #883, opened 2026-09-11
ROSAENG-66898: Enable automated UBI9 image updates

joshbranham:
Automate updates to the base UBI image in the boilerplate image, like https://github.com/openshift/boilerplate/blob/be126284e16f786622c88a303343124dddd64fbe/config/Dockerfile#L2

Then we can just cut releases using snapshots that are tied to commits that bump this, etc

Closes https://github.com/openshift/boilerplate/pull/880

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -7,9 +7,15 @@
     "group:allDigest"
   ],
   "enabledManagers": [
+    "dockerfile",
     "tekton",
     "gomod"
   ],
+  "dockerfile": {
+    "includePaths": [
+      "config/Dockerfile"
+    ]
+  },
   "tekton": {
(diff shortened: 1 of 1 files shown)


.github/renovate.json
```diff
@@ -7,9 +7,15 @@
     "group:allDigest"
   ],
   "enabledManagers": [
+    "dockerfile",
     "tekton",
     "gomod"
   ],
+  "dockerfile": {
+    "includePaths": [
+      "config/Dockerfile"
```
joshbranham, line 16: Technically if a downstream repo has a `config/Dockerfile` renovate will try and bump UBI, but they don't, nor would it be a problem.

PR #888, opened 2026-09-14
ROSAENG-67043: Add auto-release pipeline for tag pushes

joshbranham:
This adds a new `image-tag-push` pipeline that will run and make a snapshot on tags on this repo. We can then make an accompanying MR to Konflux to respond and auto-release from this new pipeline. We set the existing `image-push` to not auto release (we can also delete that once we know this works)


geowa4 (approved):
/lgtm
/approve

